PR #49, opened 2015-05-06
Preserve ability to query by entrez_id

bwalsh:
As part of integrating with the Collaborative Cancer Cloud (CCC) with CIViC, we've come to rely on the entrez_id as the mechanism to link the MAF files produced by the pipeline with the CIViC web app.

Thanks for your time.


acoffman:
Looks good, thanks for the PR!

PR #53, opened 2015-05-20
add to /api/ccc apito support batch

bwalsh:
```
 “As a Clinician or Researcher, when my CCC workflow completes, I need to see the literature and provenance for a genomic feature.”

```

The CCC workflow terminates with a MAF file.  Some of these MAF files are quite large (+150K lines)
Therefore querying CIViC for each gene is prohibitive.   This change allows the CCC workflow to query CIViC twice, once for the entrez_ids and a second time for the genomic locations.

image


acoffman:
Looks good! Thanks!

PR #54, opened 2015-05-20
ohsu data import

bwalsh:
Data extracted from CGD.

```
rake civic:import['import/ohsu-GeneSummaries.txt','import/ohsu-VariantSummaries.txt','import/ohsu-ClinActionEvidence.txt','import/ohsu-VariantGroupSummaries.txt']

```

image


malachig:
Great!  This is a significant amount of new data.  I notice that the `ClinActionEvidence.txt` is empty.  Is the plan to prime CIViC with genes and variants that are deemed clinically actionable and fill in the evidence for these clinical actions later?  @bwalsh for posterity, do you mind describing how this data was gathered here is this thread?

To date we have been driving population of CIViC from the creation of clinical evidence statements.  This means there is immediately a documented clinical relevance for each variant.  That is where a lot of the curation effort is of course. As is, this import could act as a todo list (we have several others that can be imported in that vein as well).  Curation of evidence items would happen in the interface and finer detail provenance of that information would be captured.  That would be a good thing.  

However, it should be noted that this is not how we envision contributions from domain experts using the interface to enter data (as opposed to a bulk import as you are doing).  In that scenario, the user does not enter genes or variants directly.  They enter evidence of *clinical significance* of a variant.  If the variant/gene is not there already it gets entered upon addition of the first such evidence statement.

acoffman:
@malachig I think that there are new Evidence Statements, github just isn't showing the diff:

` 1,107 additions, 0 deletions not shown `

malachig:
Ah!  Sorry, my bad. 

This is even more awesome then.  haha.  @bwalsh would still be great to have a description of how these were generated.  We are thinking about ways to mark these bulk imports that down the road we will be able to see which variants were from the OHSU import, from TGI, from Vancouver's POGS program, etc.

bwalsh:
Thanks Guys.

Below are some slides that show the overall flow.  

I do have the code that produced the files in a self contained ruby module.  While it is unobtrusive and not part of the test or production environment, it does need a secondary datasource to support the mysql data source.   I can  package that into a PR if it is useful.


image

image

malachig:
Thanks!  This is great.

malachig:
Hi @bwalsh .  We are going to take a look at these items and I think this PR should contain a discussion on any questions we have about how this is fitting in the current CIViC data model.  

First, question.  Did you import these data on top of the existing data?  Or did you create a CIViC instance with only the OHSU data.  I recommend that we try the former so that we can see head-to-head comparisons of how the OHSU bulk import compare to the MGI (our Genome Institute) bulk import.

malachig:
I went ahead and imported this data as a test on top of our existing data in a local instance of CIViC.  First off, I would like to say that this is a fantastic start and a rich source of information. Thanks!  Some of this info was not already covered in CIViC, though the numbers above have some caveats I believe, and the true degree of overlap will change as we further harmonize the import with the CIViC data model.

I observed these warnings on import:
```
Variant named MLL  REARRANGEMENT FUSION GENE for gene -1 not found!
Variant named MLL  REARRANGEMENT FUSION GENE for gene -1 not found!
Variant named MLL  REARRANGEMENT FUSION GENE for gene -1 not found!
Variant named MLL  REARRANGEMENT FUSION GENE for gene -1 not found!
Variant named MSH1 ANY MUTATION for gene -1 not found!
No disease found for DOID: N/A!
```

Based on what I describe below, I'm surprised the import process didn't throw more warnings and possibly errors in some places.

Some very early thoughts/topics for discussion:

- Something strange is happening with the variant names.  In CIViC a variant name is the name that a clinician or researcher would use.  e.g. `V600E`.  The first example from the new import that I looked at had this name `CDNA:C.1799_1800AA; CDNA:C.1799T>A; PROTEIN:P.V600E/P.VAL600GLU; BRAF V600E MISSENSE MUTATION`.  It looks like these evidence statements could be added to the existing entries for BRAF V600E.
- The variant summaries appear to be a repeat of the long variant name described above.  Over time, these need to become a human readable synthesis of the clinical significance of each variant based on the N evidence statements available.
- New evidence levels are being defined but it looks like these are not fundamentally different from those we have defined.  For example, the example I looked at above was a published case report.  This could be defined as level E according to the CIViC data model definitions.
- Trust rating is not defined for any/most entries.  This could be done as a curation task.
- Variant origin is set to 'unknown' for all entries.  This could be another curation task.  The majority of these variants are somatic (if not all of them).
- Clinical significance values are free form and not matching the enumerated list of possibilities defined by the CIViC data model.  Mapping these to CIViC notions of clinical significance could be one of the larger curation tasks needed to harmonize this bulk import with the CIViC data model.
- Citation is missing for some entries.

bwalsh:
@malachig Thanks for the detailed feed back.  

## Import into Empty CIViC
In general, we worked on top of the CIViC data.  At least one train of thought is that there should be an intersection of genes, variants, sources and drugs between the two datasets.  I have imported into an empty database separately.  I can produce a set of files that correspond to loading CGD into an empty CIViC db.  Would that be useful?

## Warning messages
Thanks for the catch on MLL & MSH1,  this was a byproduct of the underlying data from Dienstmann  v7 and suggested correction from Obi.  I've corrected this in code.

On the missing DOID, that is where the data is at this time -  We have one disease " ALK G1202R missense mutation" has a disease ("Disease") we cant match in data.bioontology.org.  I'll follow up with this in the next round of data from Dienstmann & CGD.

## Variant Names
Variant names is a challenge; both from the originating data and it's attributes.  As you know, CIViC identifies variants with name, description & internal id.

###  Variant definitions include many more attributes and subclasses [ProteinVariant, CDNAVariant,GenomicVariant]

image

### With no decoration, the UI is misleading
image

### Intermediate solution
image

### Mid-term solution
For discussion ...
* Extend CIViC Variant model for Variant types 
* Alter CIViC UX to support those types

## Variant summaries 
At this time, I don't have text to insert, I'll follow up on my side.

## Evidence Attributes
For discussion ... how to harmonize.

image

## Other Attributes
Clinical significance,  Trust rating :  as you mention, this will need to be curated.
Variant origin: I've found a solution to populate this

PR #152, opened 2016-02-09
add direct search for evidence items to advanced search

malachig:
Proposed fix for https://github.com/genome/civic-client/issues/331


malachig:
@acoffman gave the thumps up on on this so merging now.

PR #229, opened 2016-09-07
add support for civic-server dev login

ahwagner:
to be used with client-side: https://github.com/ahwagner/civic-client/tree/dev_login


config/routes.rb
```diff
@@ -3,6 +3,9 @@
   root to: 'static#index'
   scope 'api', defaults: { format: :json } do
     get '/auth/:provider/callback' => 'sessions#create'
+    if Rails.env.development?
```
acoffman: I believe this route is redundant with the one above it. It needs to exist regardless of the mode of the application.

acoffman:
Looks good to merge once the routes issue is fixed!

+2

PR #269, opened 2016-12-14
Add badge by uuid

ahwagner:
This pull request adds functionality to the new badges system. It allows admins to manage `badge claims` in the badge claim admin panel (/admin/badge_claims) or create them _en masse_ (up to 1000 per request) using the `create multiple badge claims` utility (/admin/utilities). It is designed to allow only "special" tier badges to be created in this way, although admins may create/edit single instances of badge claims associated with other badge types through the badge claim admin panel.

An authenticated user may redeem one of these claims to have the `badge` corresponding to the `badge claim` awarded to their profile.


app/models/actions/redeem_badge_claim.rb
```diff
@@ -0,0 +1,29 @@
...
+    attr_reader :badge_claim, :user
+
+    def initialize(badge_claim, user)
+      @badge_claim = badge_claim
+      @user = user
+    end
+
+    private
+    def execute
+      if @badge_claim.user.nil?
```
acoffman: Another tiny nitpick (that seems to be the theme as this PR is good overall hah), but the use of `attr_reader` sets up accessor methods for `badge_claim` and `user` and makes it so that you don't have to use the instance variable directly. In other words, if you like, you can just reference `badge_claim` directly rather than having `@`s everywhere.

app/models/badge_claim.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  validates :badge, presence: true
+
+  def redeem(user)
+    cmd = Actions::RedeemBadgeClaim.new(self, user)
+    cmd.perform
+  end
+
+  def self.create_many(badge, count)
+    BadgeClaim.create(Array.new(count) {{badge: badge}})
```
acoffman: Neat trick. I've never actually seen this before!

app/presenters/badge_claim_presenter.rb
```diff
@@ -0,0 +1,16 @@
+class BadgeClaimPresenter
+  attr_reader :claim
+
+  def initialize(claim)
+    @claim = claim
+  end
+
+  def as_json(opt = {})
+    {
+        id: claim.id,
```
acoffman: In the migration, you add timestamps to this (updated_at and created_at), it may make sense to display them in the JSON.

db/migrate/20161214033216_enable_uuid_extension.rb
```diff
@@ -0,0 +1,11 @@
+class EnableUuidExtension < ActiveRecord::Migration
+  def change
+    enable_extension 'uuid-ossp'
```
acoffman: This is definitely the "correct" way to do this. Unfortunately, I'm not confident that the rails user (`www-data`) in production has sufficient permissions to enable postgres extensions unilaterally. We could potentially enable it in puppet, or we could work around it in a slightly different manner.

You could use a normal integer primary key and then create a `text` type column for the UUID. Then in an activerecord callback in your `BadgeClaim` object (probably `after_create`) you could do something like

```ruby
after_create :init_uuid

def init_uuid
  unless self.redemption_code.present?
    self.redemption_code = SecureRandom.uuid
    self.save
  end
end
```

app/admin/utilities.rb
```diff
@@ -75,6 +75,24 @@
         end
       end
       column do
+        panel 'Create multiple badge claims' do
+          all_special_badges = Badge.where(tier: 'special').sort_by(&:created_at).reverse
```
ahwagner: Thanks for the tip! I agree, it makes more sense to do this database-side. I'll change it.

app/controllers/badge_claim_controller.rb
```diff
@@ -0,0 +1,22 @@
+class BadgeClaimController < ApplicationController
+
+  def create
+    @claim = BadgeClaim.new
+  end
+
+  def delete
+    @claim = BadgeClaim.find(params[:claim_id])
+    @claim.delete
```
ahwagner: Yeah, I had originally intended to make routes for these endpoints, but ultimately ended up going the admin panel route instead (it made more sense to me, since non-admin users ultimately would never use these endpoints).

Good catch, I didn't mean to leave this code behind. I'll remove it.

app/controllers/badge_claim_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def redeem
+    @claim = BadgeClaim.find(params[:claim_id])
```
ahwagner: Yeah, I think I was being overly cautious here. I agree, instance variables aren't needed here. Changing it.

app/models/actions/redeem_badge_claim.rb
ahwagner: Cleaned this up too. Looks way prettier without all those `@`s.

app/models/badge_claim.rb
ahwagner: Thanks! There are advantages to stumbling around Rails without knowing much about common practice.

db/migrate/20161214033216_enable_uuid_extension.rb
ahwagner: I'm glad that I started "correct", although I think that avoiding a puppet workaround is best. I'll implement your suggestion here.

ahwagner:
😁

PR #280, opened 2017-01-25
Added ability to search integer fields for null values

agraubert:
Backend change for genome/civic-client#648

It's really similar to the is_empty comparison, but TRIM() can't take integers and I couldn't think of a way to modify the existing comparison to keep it's original function while also accepting integer fields as input


acoffman:
Looks good to me!

PR #281, opened 2017-01-25
Fixed issue with IS NOT search in clinical significance

agraubert:
Searching evidence items for 'Clinical Significance is not ...' replies with HTTP 500

I can't link directly to the search but you can go [here](https://civic.genome.wustl.edu/#/search/evidence/a365e757-c530-4350-82f5-5c2cfa2928b3) and change the condition from `is` to `is not`


acoffman:
Also looks good, nice catch!

PR #289, opened 2017-02-03
Enabled searching evidence by pmc_id

agraubert:
@acoffman I know you mentioned scrapers in genome/civic-client#660, so I don't know if you think there's more functionality that needs to be added with regards to pmcid's.  This was just the base change to have the server respond to the pmcid searches


acoffman:
Looks good to me! @kkrysiak was right, we're already scraping PMCID as a proxy for "open access."

kkrysiak:
If you could include an "is empty" and possibly "has/is not empty" as search options, that would be useful.

agraubert:
Okay, I'm adding the server change for that in #291 and the front end change to the existing PR genome/civic-client#677

PR #294, opened 2017-02-17
Updated backend to support client HTML5 mode

agraubert:
Updates the backend to support the new HTML5 mode URLs in the frontend introduced in genome/civic-client#693.  Since the changes to markdown entity linking will produce new-format URLs, this shouldn't be merged until the frontend is.  To minimize the window with broken links, we should take a new frontend build whenever this gets merged.

Added a new task to generate a page of links to all database entities, to assist Google's indexing.  This task takes about 2s to run, and is scheduled to run nightly, but will need to be scheduled initially from the console with: `GenerateList.perform_later`


acoffman:
👍

PR #317, opened 2017-05-15
update version

ahwagner:
This is needed to update to the newest civic data dump. We should probably evaluate whether or not we actually want to keep this check for the civic:load task, especially if we start generating sanitized dumps on a regular basis.


ahwagner:
Sounds good. I'm closing the PR and opened #318 so this can be addressed once the Puppet deploy code is updated.

Issue #318, opened 2017-05-15
Local data.sql loading


ahwagner:
The civic database is checked for "version" on deploy, as a holdover from an earlier method for importing data into the database. This should be addressed in the puppet deploy code, and db versioning should no longer be checked as part of the `civic:load` task.

See https://github.com/genome/civic-server/pull/317#pullrequestreview-38215835 for prior discussion on the issue.

PR #326, opened 2017-05-25
stop-gap measure for #318 allows forced db updates from rake task

ahwagner:
Allows the use of `rake civic:load[force]` to disregard version in database update. Does not change behavior of `rake civic:load` without options.


ahwagner:
🎉

PR #350, opened 2017-08-07
First draft for assertions

susannasiebert:


susannasiebert:
@jmcmichael The endpoint for all assertions related to a variant can be found at `/api/variants/:id/assertions` and the details endpoint for a single assertion is at `/api/assertions/:id`.

app/models/assertion.rb
```diff
@@ -0,0 +1,95 @@
...
+    class_name: Event
+
+  enum evidence_type: Constants::EVIDENCE_TYPES
+  enum nccn_guideline: Constants::NCCN_GUIDELINES
+  enum amp_level: Constants::AMP_LEVELS
+  enum acmg_level: Constants::ACMG_LEVELS
+  enum clinical_significance: Constants::ASSERTIONS_CLINICAL_SIGNIFICANCES
+
+  def self.index_scope
+    all
```
acoffman: This should likely include all the tables that are loaded in the `AssertionIndexPresenter` which looks like `Disease`, `Regulatory Agencies`, `Country`, `Evidence Item Count`, etc. Basically, you'll want to make sure that the index view generates a fixed number of queries rather than a number of queries proportional to the number of items loaded.

app/models/assertion.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def self.view_scope
+    eager_load(:acmg_codes, evidence_items: [:disease, :source, :drugs, :variant])
```
acoffman: Similarly here, I'd double to check make sure you're loading everything you should be

app/models/assertion.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def name
+    "AID#{self.id}"
+  end
+
+  def parent_subscriables
+    self.evidence_items
```
acoffman: This is a little odd as these don't really cleanly fit into our parent/child hierarchy for subscriptions and notifications. We should give some thought as to how we want that to work in this case as it seems like assertions can sit at multiple levels of the hierarchy.

susannasiebert: Agreed. Would it work to make the parent_subscribables the variant ad the child_subscribales the evidence items and sort of ignoring the whole indirectly related aspect?

Also, I should fix that typo or this won't work at all.

PR #351, opened 2017-08-09
Scoped hierarchical event feed endpoint

acoffman:
This allows for querying a specific entity (say a variant) and getting a
custom, paginated, event feed for the entity in question as well as all
child entities.

The new endpoint exists at `/api/events/(genes|variants|evidence_items|etc)/:id`


acoffman:
@susannasiebert I'm interested in feedback about the new `EventHierarchy` class. Previously, each model that was `subscribable` optionally implemented `parent_subscribables`. This was because a change to an entity was intended to notify users that had subscribed to parent entities.

Creating these scoped event feeds now requires traversal of the hierarchy in the opposite direction as entities now need to be aware of their children in addition to their parents. My first pass involved just adding `def child_subscribables` to each class in question, but that started to feel messy. And It seemed a little weird that a `Gene` or `Variant` would contain logic relating to our event/notification propagation hierarchy.  Thus was born the `EventHierarchy` class which contains all that information in one place. On the one hand this is cleaner, on the other hand, this one class is pretty ugly... so I'm not really sure if its better or worse.

There's some other weird bits in this code, so I'm going to leave some inline comments explaining parts of it.

app/models/scoped_event_feed.rb
```diff
@@ -0,0 +1,38 @@
...
+    @event_subjects = EventHierarchy.new(root).children
+    @page = page
+    @per = per
+  end
+
+  def events
+    if @events
+      @events
+    else
+      event_ids = event_subjects.group_by { |x| x.class.to_s }.inject([]) do |ids, (type, vals)|
```
acoffman: We get a list of `Event` subjects from our hierarchy, but unfortunately, the "or" methods available in ActiveRecord don't seem to play nicely with spanning multiple types in our polymorphic foreign key. (Ie: I can't just do Event.where(subject: subjects)` as it doesn't generate the SQL we'd want. 

Instead we group by the class of the subject (`Gene`, `Variant`, etc) and run one query for each type, but just pull back the Event `id` rather than the full object. So we'll have one query in the form of `select events.id from events where subject_type = 'Variant' and subject_id in (1, 2, 3);` for each type.

We aggregate these ids into a list of event ids and then perform our "real" query below, pulling back all the data we want, paginating, etc.

app/models/scoped_event_feed.rb
```diff
@@ -0,0 +1,38 @@
...
+        .page(page)
+        .per(per)
+    end
+  end
+
+  def comments
+    if @comments
+      @comments
+    else
+      comment_ids = events.select { |e| e.state_params['comment'].present? }
```
acoffman: Unfortunately the `subject` of a `commented on` event is the thing being commented on, not the comment itself. That means we're doing some gymnastics here to grab all the comments mentioned in any of the events in our feed so that @jmcmichael can display them inline (a la Github). We're basically just building up a hash of comment_id -> comment object so we can look them up quickly when generating the json.

app/presenters/scoped_events_presenter.rb
```diff
@@ -0,0 +1,34 @@
...
+    {
+      _meta: PaginationPresenter.new(event_feed.events, [], request),
+      records: event_feed.events.map { |e| event_with_comments(e) }
+    }
+  end
+
+  private
+  def event_with_comments(event)
+    event_json = EventPresenter.new(event).as_json
+    if event.action == 'commented'
```
acoffman: For "commented on" actions, include the comment in addition to all the normal event fields.

app/models/scoped_event_feed.rb
acoffman: After talking to @susannasiebert we're going to try and generate some parameterized SQL here and see if its any more clear.

acoffman:
@susannasiebert, I updated the query generation to create a parameterized SQL string instead of the groups of event ids. Let me know if you think this is more clear!

susannasiebert:
@acoffman It's more wordy but also more readable.

acoffman:
Rails 5 actually allows for the query to work in ActiveRecord in the more natural way you'd expect (the way I'd mentioned wanting to write it [here](https://github.com/griffithlab/civic-server/pull/351#discussion_r132306565) so I'm going to push the PR to upgrade to Rails 5 this weekend, and then merge that into this one and rewrite the query to not suck!

PR #352, opened 2017-08-10
Add endpoint for all evidence items in a variant group

susannasiebert:
This implements the endpoint necessary for https://github.com/griffithlab/civic-client/issues/800.

The endpoint can be found at `api/variant_groups/:id/evidence_items`.

### DIFF
--- a/app/controllers/evidence_items_controller.rb
+++ b/app/controllers/evidence_items_controller.rb
@@ -2,7 +2,7 @@ class EvidenceItemsController < ApplicationController
   include WithComment
   include WithSoftDeletion
 
-  actions_without_auth :index, :show, :variant_index
+  actions_without_auth :index, :show, :variant_index, :variant_group_index
 
   def index
     items = EvidenceItem.index_scope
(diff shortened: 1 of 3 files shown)


app/models/evidence_item.rb
```diff
@@ -74,6 +74,10 @@ def self.advanced_search_scope
     eager_load(:submitter, :disease, :source, :drugs, :open_changes, variant: [:gene, :variant_aliases])
   end
 
+  def self.variant_group_scope
+    eager_load(variant: [:variant_groups])
```
acoffman: You'll likely want to include the `index_scope` here as well as the joins you need for the variant group query. In theory, you'll know you've gotten it correct if you look at the server logs and see only one query (or a handful) rather than 1 per evidence item. 

Making `variant_group_scope` something like ` eager_load(:disease, :source, :drugs, :open_changes, variant: [:variant_groups])` would probably be a good place to start!

app/controllers/evidence_items_controller.rb
```diff
@@ -34,6 +34,21 @@ def variant_index
     )
   end
 
+  def variant_group_index
+    variants = EvidenceItem.variant_group_scope
+      .order('evidence_items.id asc')
+      .page(params[:page].to_i)
+      .per(params[:count].to_i)
+      .where(variant_groups: { id: params[:variant_id] })
```
acoffman: Does this work as written? I would expect it'd need to be `variant_group_id`

susannasiebert: Good catch. I basically copied this method from the variants controller, which also contains this bug. I made a PR to fix the variants controller (#356).

app/models/evidence_item.rb
susannasiebert: I made this change but I'm still seeing a few queries when I hit this endpoint in the server (same with #353). Not sure if that is to be expected.

PR #353, opened 2017-08-10
Add endpoint for all evidence items of an organization

susannasiebert:
These are evidence items that were submitted by a user of that organization.

The new endpoint can be found at `api/organizations/:id/evidence_items`.

### DIFF
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -45,7 +45,7 @@ class OrganizationsController < ApplicationController
   def evidence_items
     evidence_items = EvidenceItem.order('evidence_items.id DESC')
       .eager_load(:submitter)
-      .where('users.id': user_ids)
+      .where('users.id' => user_ids)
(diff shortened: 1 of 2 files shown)


app/controllers/organizations_controller.rb
```diff
@@ -43,4 +41,23 @@ def events
       PaginationPresenter
     )
   end
+
+  def evidence_items
+    evidence_items = EvidenceItem.order('evidence_items.id DESC')
+      .eager_load(:submitter)
+      .where('users.id': user_ids)
```
acoffman: Does this work as written? I'd think you'd need `'users.id' => user_ids` as `thing: other_thing` is shorthand for `:thing => other_thing` and `users.id` isn't a symbol, its a string.

susannasiebert: It returns the same number of results so I think it works either way? I changed the clause, just to be sure.

PR #354, opened 2017-08-11
WIP: New endpoint for community stats

susannasiebert:
This implements the endpoints needed for https://github.com/griffithlab/civic-client/issues/788.

The new endpoint can be found at `api/stats/community`.


susannasiebert:
We probably want to keep the list of countries as a new database table in order to constrain what the user can enter and make the `country` field on the database a foreign key to that new table.

app/controllers/stats_controller.rb
```diff
@@ -29,6 +29,17 @@ def current_user_stats
     render json: stats_for_user(current_user)
   end
 
+  def community
+    stats = {
+      badge_user_count: Badge.all.each_with_object({}) {|b, h| h[b.name] = b.badge_awards_count_per_tier },
+      role_user_count: User.all.each_with_object(Hash.new(0)) {|u, h| h[u.role] += 1},
```
acoffman: This all looks like its correct and works, so this is more general advice than a correction! There's a few things I think we can do with this stats endpoint in general to lean on the database more and do less in memory.
### Badge Count:

The `badge_awards_count_per_tier` method is probably redundant with just a simple sql GROUP BY:
``` ruby
Badge.all.each_with_object({}) { |badge, h| h[badge.display_name] = badge.badge_awards.group(:tier).count }
```

### Roles:
`User.group(:role).count` will just do a `GROUP BY` in SQL and you'll get a hash out of role -> count. The caveat though is that roles are an enum so the actual role in the db will be an int. ActiveRecord provides a handy hash of the values to look them up though. So something like:
```ruby
User.group(:role)
  .count.map {|role, count| [User.roles.invert[role], count] }
  .to_h
```

Will put the majority of the work on postgres instead of in memory in ruby land. Furthermore (and this may be a question for a meeting) it might make sense to inner join to the events table so we only count users that have actually done something (though maybe we want to count them all?). If so, you could do that like this:
```ruby
User.joins(:events)
  .group(:role)
  .uniq
  .count.map {|role, count| [User.roles.invert[role], count] }
  .to_h
```

You can imagine handling the other stats similarly as well if you wanted. On top of all of that, we should probably cache these results as they could potentially get expensive to calculate over time. `Rails.cache.fetch` provides a convenient way to do this.

Basically it takes a key (its a key-value store), an expiration time, and a block. It looks up the key, if its there, it returns the result, otherwise it runs the block, puts the result in the cache, and returns it. So something like:

```ruby
Rails.cache.fetch("community-stats-summary", expires_in: 10.minutes) do
{
 #big stats hash here
}
end
```

Would work well!

susannasiebert:
I refactored the calculation of the stats like Adam suggested. The caveat is, that certain stats will filter out `nil` values which are, for example, users that haven't specified a organization, area of expertise, or country. We should decide if we'd rather include them as an "unknown" bin.

malachig:
It is probably okay to limit these stats to those people who have defined the field...

susannasiebert:
This PR probably doesn't include all of the desired stats we ultimately want to display but I will merge it in for now and make new PRs once we have a better idea what we'd like to add or change.

Issue #141, opened 2016-01-07
Develop a variant and gene 'clinical relevance score' that can be used to sort variant and gene browse grids


malachig:
Currently when looking at the browse variant or gene pages we can sort by number of evidence items.  This is nice, but it is possible for a gene/variant to have a large number of evidence items that are all low evidence levels and/or low star rating.

We should calculate a 'clinical relevance score' for each variant and gene.  A single integer value that is higher with more evidence items, but is weighted to give a higher score for variants with high evidence levels and star ratings.

We should use this issue to discuss how it might be calculated.

malachig:
Potentially important factors:

From evidence statements
- Evidence level (A > B > C > D > E)
- Star rating (5 > 4 > 3 > 2 > 1)
- Evidence direction (don't count or penalize for negative statements)
- Evidence type?

Things that are aggregate across the set of evidence items? e.g.
- The variant has evidence items for multiple cancer subtypes?
- The evidence is spread across multiple journals/publications?
- The evidence was contributed by multiple users?
- Others

kkrysiak:
Other aggregates:
- The number of drugs associated with predictive evidence?
- Incorporate variant groups somehow?

susannasiebert:
We need to add the score as a search metric for variants and evidence items.

PR #364, opened 2017-11-02
Initial implementation of cached variant scores

acoffman:
This will require some further testing on my end - but it looks like master of the client repo is perhaps broken at the moment.

This is an initial implementation of a cached variant score for the civic panel. It will recalculate the score in the background and save it on the variant each time a new evidence item is accepted or a change to an evidence item is accepted. We can initially populate all the scores with something like

```
Variant.all.each do |v|
  UpdateVariantScore.new.perform(v)
end
```

There are a few gaps here. If something is done directly in the console or via the admin this will not trigger. Likewise for a scripted bulk import. We could close that gap by making this a literal callback on evidence item in `after_save` but that seems messier to me and would fire every time we saved an evidence item rather than just in these specific cases. I'm open to suggestion though if you have any feedback @susannasiebert.

Additionally, before this PR is merged we'll likely want to start using this column to generate the panel report TSVs instead of calculating on the fly which should improve performance significantly.


susannasiebert:
This all look good to me. I agree that we don't want this to trigger for every change. The same caveats are true for other things as well (e.g. badges and events) so I don't think this needs any special handling.

RE the client: I believe Josh pushed his assertions changes directly to the client master and since the assertions PR on the server isn't merged yet, there are conflicts. I wonder if we need to "reset" the client and put the assertions changes into a branch and rollback master to the last released commit. Unfortunately, Josh is probably out for a while so I won't be able to discuss this with him. It would be nice to get this feature out without also releasing assertions prematurely since those still need some work.

susannasiebert:
I pushed a commit to have the panel report use the new column. However, this would break things until the column is backfilled. Should we change the migration to "backfill" the column with a default value?

susannasiebert:
(This PR is related to #141)

Issue #363, opened 2017-10-30
Pull in data from AlleleRegistry


kkrysiak:
When we can match, we should link to Allele Registry data
http://reg.clinicalgenome.org/redmine/projects/registry/genboree_registry/landing

They allow for matching using HGVS or chr, start, stop, ref, var. I would recommend the latter. 

Ultimately we would want to pull in additional information but at first we would want to pull in the ID and link out to the allele registry. This would be a good collaborative effort with ClinGen. We can register alleles that don't match.

kkrysiak:
The 13 is the version. You can see on ClinVar where the 14 is associated with GRCH38. https://www.ncbi.nlm.nih.gov/clinvar/variation/13961/

Unfortunately, these numbers are not universal. As in 13 != GRCH37. They iterate the version as updates are made. https://www.ncbi.nlm.nih.gov/clinvar/variation/12582/

malachig:
To elaborate a bit further.

To be a valid HGVS expression, the sequence accession before the ":" should unambiguously refer to a **sequence** (e.g. DNA, RNA, protein).  The HGVS expressions we are using for MyVariant.info do not follow this rule.  "chr7" could refer to many sequences. "NC_000007" is a GenBank accession that refers to a submission record.  But to get to an unambiguous sequence you also need a version number.  "NC_000007.13" refers to a single unambiguous sequence.  It seems that these version numbers are incremented if the sequence changes, or when a new genome build is released.

For a particular version of the genome (e.g. build37), we can probably safely assume a simple mapping between the "chr7" style representations, and the "NC_000007.13".  Adam did this when he automatically created HGVS expressions at the genome level for all simple variants.  These were submitted in bulk to CIViC and then reviewed by editors.  We could follow the same approach here.

malachig:
![accessions](https://user-images.githubusercontent.com/856728/32507064-0b2392d6-c3ac-11e7-9162-389c838be1b2.png)

Allele registry has a handy reference table for these accessions.

susannasiebert:
Thank you. That's really helpful.

obigriffith:
You can also find this through the NCBI assembly pages. Most current GRCh37 is at: https://www.ncbi.nlm.nih.gov/assembly/GCF_000001405.25/

most current GRCh38 is at: 
https://www.ncbi.nlm.nih.gov/assembly/1132361/

kkrysiak:
Basic information is pulled in. I'm going to create a new issue for expanding that capability.

PR #365, opened 2017-11-07
Add proxy for allele registry

susannasiebert:
I believe that this should be what's needed for #363  on the server side to get the Allele Registry info displayed.

### DIFF
--- a/app/models/variant.rb
+++ b/app/models/variant.rb
@@ -69,11 +69,7 @@ class Variant < ActiveRecord::Base
 
   def allele_registry_id
     response = AlleleRegistry.new(id).response
-    if response.instance_of?(String)
-      JSON.parse(response)['@id'].split('/')[-1]
-    else
-      nil
-    end
+    JSON.parse(response)['@id'].split('/')[-1] rescue nil
   end
 
   def state_params
--- a/app/models/variant.rb
+++ b/app/models/variant.rb
@@ -70,7 +70,7 @@ class Variant < ActiveRecord::Base
   def allele_registry_id
     response = AlleleRegistry.new(id).response
     if response.instance_of?(String)
-      JSON.parse(response)['@id']
+      JSON.parse(response)['@id'].split('/')[-1]
     else
       nil
     end
--- a/app/models/variant.rb
+++ b/app/models/variant.rb
@@ -67,6 +67,15 @@ class Variant < ActiveRecord::Base
     name
   end
 
+  def allele_registry_id
+    response = AlleleRegistry.new(id).response
+    if response.instance_of?(String)
+      JSON.parse(response)['@id']
+    else
+      nil
+    end
+  end
+
   def state_params
     {
       gene: {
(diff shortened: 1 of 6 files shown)


susannasiebert:
The new endpoint for this is at `variants/:id/allele_registry_proxy`

susannasiebert:
The linkouts should be to the following URL: `http://reg.genome.network/allele/:ca_id.html`. The CA ID is embedded in the `@id` field of the above endpoint.

susannasiebert:
I just pushed another commit to this PR to change the API URL to the actual production server instead of using the test server. As a result the `@id` field now returns a production server URL (e.g., `http://reg.genome.network/allele/CA123643`). The linkout URL can now be more easily constructed by appending `.html` to it.

susannasiebert:
@jmcmichael I added a new key to the variant show endpoint. `allele_registry_id` will give you the @id value of the allele registry API call, e.g. `http://reg.genome.network/allele/CA123643` where `CA123643` is the allele registry ID we want to display and `http://reg.genome.network/allele/CA123643.html` is the link target.

app/presenters/variant_detail_presenter.rb
```diff
@@ -9,6 +9,7 @@ def as_json(opts = {})
         clinvar_entries: variant.clinvar_entries.map(&:clinvar_id),
         lifecycle_actions: LifecyclePresenter.new(variant),
         sources: variant.sources.map { |s| SourcePresenter.new(s) },
+        allele_registry_id: variant.allele_registry_id,
```
acoffman: If this is going to be on every variant response, it may make sense to add it as a field in the database and lazily populate it as needed rather than fetching it every 24 hrs as we're doing here.

app/models/variant.rb
```diff
@@ -67,6 +67,15 @@ def display_name
     name
   end
 
+  def allele_registry_id
+    response = AlleleRegistry.new(id).response
+    if response.instance_of?(String)
+      JSON.parse(response)['@id'].split('/')[-1]
```
acoffman: This will die if we don't get a response or `@id` isn't present. To preserve the current behavior in case of this sort of failure you could do something like `JSON.parse(response)['@id'].split('/')[-1] rescue nil` which will just catch all a exceptions and return nil if one is thrown.

susannasiebert: I'll add that real quick.

app/presenters/variant_detail_presenter.rb
susannasiebert: Ultimately, we want to display more information from the response (similar to MyVariantInfo) but we wanted this minimal feature out in time for Obi's presentation on Friday. It might still make sense to store this as a field in the database. We'd need to make this a callback to save when the variant gets approved and update should the variant coordinates change.

acoffman: Cool, makes sense! Generally for external sources like this, we don't actually return any of it as part of the civic record (mygeneinfo and myvariantinfo as examples) they just go as a separate request/record, but if we have a good reason to do a hybrid style record in this case (civic variant + external source) this works fine.

susannasiebert: Yeah, mainly we just wanted to avoid having to make another request to one of our API endpoints on the client side.

acoffman: Yeah, in general I like keeping the stuff that isn't part of the civic record in a different record, but I can see where we may want to try and cut down on requests for efficiency purposes. Perhaps we need a `/variants/:id` for the civic record and then some sort of `/variants/:id/complete` or `/variants/:id/frontend_display` or something that stitches together the disparate records in a single request.

The other consideration is that with one big slow(er) request, we can't render anything until the request is completed. With multiple smaller requests, we can be doing the database work in parallel and render different components as the responses come back, so its not always worse to have multiple requests.

PR #367, opened 2017-11-15
fixes issue with /variants_status endpoint

jmcmichael:
/variants_status was returning the entire gene record, reverting griffithlab/civic-server@e5010bd fixed the problem.


susannasiebert:
+2

acoffman:
We need the fix that the faulty commit introduced, but yeah, this is no good as a regression. I'll take a look at it, but definitely go ahead and merge the fix!

PR #384, opened 2017-12-22
Remove hardcoded panel types and type-specific presenters.

acoffman:


susannasiebert:
This looks good to me. I believe we also need to update the values in the pipeline_types table.

acoffman:
@susannasiebert, we definitely do! :)

Issue #390, opened 2018-01-18
I broke flags


kkrysiak:
I merged E384* and E384Ter in the admin interface, except I did this before I resolved the open flag on E384Ter. Now the Open Flag page is broken and there is an outstanding Open Flag on E384Ter.

The issue the flag caused was resolved by the merge, but I didn't resolve the flag prior to the merge.

acoffman:
The immediate issue is fixed and I commented on the PR with the fix going forward. Closing this!

PR #391, opened 2018-01-18
Add handling of flags during variant merging

susannasiebert:
Addresses #390 

This PR will change the behavior of merging variants to move flags of the variant to be deleted to the other variant and prepend ‘(Merged from [variant name])’ to the flag comment.

We still need to fix the production instance by moving the flag manually or resolving it.

### DIFF
--- a/app/jobs/merge_variants.rb
+++ b/app/jobs/merge_variants.rb
@@ -8,6 +8,7 @@ class MergeVariants < ActiveJob::Base
     ActiveRecord::Base.transaction do
       transfer_evidence_items
       remove_suggested_changes
+      transfer_flags
       transfer_subcriptions
       remove_variant
     end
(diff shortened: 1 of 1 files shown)


acoffman:
This looks good to go ahead and merge and it would "mostly" work, the one thing that I see missing is that `Event`s  generated by the flagging will still point to the old variant. Basically you'd want to update those by grabbing them, re-calling `event.state_params = event.subject.state_params; event.save` or similar. It appears the merge task actually misses re-parenting events in general so that's not particular to your fix, we should maybe make a separate issue to make sure we re-parent all the events (and their notifications) when we do a merge, otherwise clicking the events won't work anymore. (Though the events would now also be kind of lies in that they'll say "Adam did X to variant Y" when in fact I maybe did that to variant Z which got merged into Y later... so maybe we want to think about that)

I'd say go ahead and merge this fix for now.

PR #428, opened 2018-05-23
Passenger restart on deploy

ahwagner:
This commit updates `capistrano-passenger` and adds the `app` role to deploy scripts to facilitate passenger restart on deploy.

With this change, we should not need to restart Apache to get the app to update.


ahwagner:
Also includes 516b2e4, which helps indicate whether an open tab is for staging or production.

PR #441, opened 2018-10-02
Initialize the civic_actionability_score to 0 when a variant is first created

### LINKED ISSUE #424: civic_actionability_score doesn't always get set
Some variants in the production database don't have a actionability score set (I had to backfill them manually). I suspect this happens for new variants. If that is the case, we just need to caluculate this field whenever a new variant gets created.

susannasiebert:
Otherwise this attribute will be nil until an evidence item for the variant is accepted, which will result in the panel API endpoints failing because we are comparing the minimum score cutoff to nil values.

Closes #424

### DIFF
--- a/app/models/variant.rb
+++ b/app/models/variant.rb
@@ -26,7 +26,7 @@ class Variant < ActiveRecord::Base
   after_initialize :init
 
   def init
-    self.civic_actionability_score = 0
+    self.civic_actionability_score ||= 0 if self.has_attribute? :civic_actionability_score
   end
 
   def self.index_scope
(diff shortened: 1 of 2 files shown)


susannasiebert:
@tmooney, good point. I made a callback instead that sets this value after_initialize.

app/models/variant.rb
```diff
@@ -23,6 +23,12 @@ class Variant < ActiveRecord::Base
 
   enum reference_build: [:GRCh38, :GRCh37, :NCBI36]
 
+  after_initialize :init
+
+  def init
+    self.civic_actionability_score = 0
```
tmooney: Would someone ever want to set an explicit value during creation that we need to avoid stomping?  (Don't see any in the codebase currently.)  If so `||=` would go well here.

PR #448, opened 2018-10-18
Enable searching of variants by allele registry id

susannasiebert:
https://github.com/griffithlab/civic-client/issues/879

The direct link is `civicdb.org/links/allele_registry/<allele registry id>`


susannasiebert:
Note to self: once this is live, I need to update the API docs to include this link.

PR #454, opened 2018-10-29
Allow organizations to have parent organizations

susannasiebert:
Closes https://github.com/griffithlab/civic-client/issues/965

This change allows you to add a "parent" organization for an organization. An organization then has `groups`, which are all organizations it is a direct parent of. We can then assign users to these new "group" organizations instead of their parent organization in the admin panel.

`all_users` returns a list of an organization's users and all its groups' users so with that change and the change to the `stats_hash` an organization's stats will include all the counts for its users and its groups' users.

### DIFF
--- a/app/admin/organization.rb
+++ b/app/admin/organization.rb
@@ -1,11 +1,12 @@
 ActiveAdmin.register Organization do
   menu :priority => 99
 
-  permit_params :name, :url, :description, :profile_image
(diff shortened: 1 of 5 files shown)


susannasiebert:
I don't think we'll want arbitrarily deep nesting of the organizations but there is currently nothing implemented to stop us from doing so. I would need to do some more research to see if it is possible to limit this somehow. Maybe @acoffman has some ideas how to do this.

PR #465, opened 2018-12-06
WIP: Normalize Drugs

susannasiebert:
This PR normalizes our drug entities to NCI Thesaurus terms. The behavior will be similar to diseases in that the users will use a unique ID (NCIT ID) to search for the drug/treatment of choice but would be given the choice to click a checkmark if a NCIT ID for their chosen term isn't available.

TODO:
- Our existing drug entities need to be normalized. The `misc_scripts/backfill_drugs.rb` script will do part of this process by querying NCIT for our drug names and backfilling the ncit_id field, as well as updating our drug name to match the NCIT term. Some manual curation will be needed for our entities that don't have a direct mapping to NCIT terms. For this we need to ensure that our spreadsheet is correct (https://docs.google.com/spreadsheets/d/1FL_Mim29q5Osloazj7rbNFoD4sdga247GE__qN1b4yc/edit#gid=14227486).
- There are probably some change that need to be made for submitting/editing evidence items both on the client and the server

This is for issue https://github.com/griffithlab/civic-client/issues/977.


lib/importer/nci_thesaurus_mirror.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    def valid_entry?(entry)
+      semantic_types = semantic_types(entry)
+      (entry['id'].present? && entry['name'].present? && entry.respond_to?(:name) && entry.name == 'Term' &&
+        (semantic_types & ['Pharmacologic Substance', 'Pharmacological Substance', 'Clinical Drug', 'Therapeutic or Preventive Procedure']).length > 0 &&
+        (semantic_types & ['Obsolete_Concept']).length == 0)
+    end
+
+    def semantic_types(entry)
+        entry['property_value'].select{|s| s.start_with? 'NCIT:P106'}.map{|t| t.rpartition(' ')[0].split(' ', 2)[1].gsub!('"', '')}
```
acoffman: This line looked a little complicated so I started trying to work through it piece by piece. It does seem to work as is, but I wonder if it would be a little more clear with a regex in this case. Up to you if you think its better or not. I tested the regex itself on the file so I think it works, but the rest isn't fully tested.

```
def semantic_types
  matcher = /^NCIT:P106 "(?<semantic_type>.+)"/
  entry['property_values'].map { |s| s.match(matcher) }.compact.map { |s| s[:semantic_type] }
end
```
`match(some_regex) returns a match object if it matched, `nil` otherwise, `compact` removes all the `nils` from the list, and then `match_object[name of capture group]` gets the value out. (You don't have to use named capture groups and then you can just do `match_object[0]` or whatever, but I like naming them.

Up to you though, either way, you probably want to avoid `gsub!` (and most `!`) methods in general as their return values can bit you sometimes. Things with `!` modify the value inline instead of returning a copy with the modifications done. The non-`!` versions will always return a copy of the original object with the changes done (or not if they didn't apply). The `!`  versions however will often return not the object, but some sort of status indicator. For instance `gsub!` returns the string if it made any changes, but `nil` if it didn't, which can be a little surprising.

lib/importer/nci_thesaurus_mirror.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    def create_object_from_entry(entry)
+      name = Drug.capitalize_name(entry['name'])
+      ncit_id = entry['id'].sub('NCIT:', '')
+      drug = ::Drug.where(ncit_id: ncit_id).first_or_initialize
+      drug.name = name
+      synonyms = process_synonyms(entry['synonym']).uniq
+      synonyms.each do |syn|
+        drug_alias = ::DrugAlias.where(name: syn).first_or_create
+        if not drug.drug_aliases.map{|a| a.name.downcase}.include? syn.downcase and not syn.downcase.eql? drug.name.downcase
```
acoffman: Generally speaking you'll want to use `!` instead of `not` and `==` instead of `.eql?` for this sort of thing. `not` has a weird and different operator precedence than you might expect and can sometimes behave in ways that are odd.

misc_scripts/backfill_drugs.rb
```diff
@@ -0,0 +1,40 @@
...
+    response['collection'].each do |r|
+      if r['prefLabel'].downcase == drug.name.downcase
+        match = r
+        break
+      end
+    end
+  else
+    next
+  end
+  if not match.nil?
```
acoffman: Same comment about `not` here, also could this just be `if match` or `if match.present?` instead of the double negative?

app/controllers/drugs_controller.rb
```diff
@@ -29,7 +29,12 @@ def name_suggestion
     if params[:q].blank?
       render json:  {errors: ['Must specify a query with parameter q']}, status: :bad_request
     else
-      render json: DrugNameSuggestion.suggestions_for_name(params[:q]), status: :ok
```
acoffman: Is `DrugNameSuggestion` used anywhere else with this gone? If not, we should delete it.

PR #469, opened 2018-12-06
Update dependencies

acoffman:
Seems to work locally but we'll want to test more before merging.


acoffman:
This PR should address all security alerts on the repo as well.

acoffman:
When we merge this, we will also likely have to update the ruby version running on the prod and staging boxes. We may as well do dgidb (and docm?) at the same time if the updates don't require major code changes (they didn't for civic). Any thoughts @ahwagner? 

Also as an aside, people who have a local copy of civic that haven't updated their ruby version since the initial setup will likely need to update their local ruby too....

PR #473, opened 2019-01-09
Include source type in source display name

acoffman:
This will cause it to appear in the diffs for revisions.
Closes griffithlab/civic-client#1020

Here's how it looks: 
screen shot 2019-01-09 at 10 40 14 am

### DIFF
--- a/app/models/source.rb
+++ b/app/models/source.rb
@@ -37,7 +37,18 @@ class Source < ActiveRecord::Base
   end
 
   def display_name
-    description
+    "#{self.description} - (#{source_type_description})"
+  end
+
+  def source_type_description
+    case self.source_type
+    when 'ASCO'
+      'ASCO Abstract'
+    when 'PubMed'
+      'PubMed'
+    else
+      ''
+    end
   end
 
   def source_url


app/models/source.rb
```diff
@@ -37,7 +37,18 @@ def self.datatable_scope
   end
 
   def display_name
-    description
+    "#{self.description} - (#{source_type_description})"
```
susannasiebert: I think the display_name might be used in other places as well so we should make sure that we want to display the source type there as well.

acoffman: Yeah, I was trying to figure that out - we have both `description` and `display_name`, we should probably take a quick inventory of what's used where. I did check the main evidence item view and it did not change it there (which I think is the correct behavior).

susannasiebert: Yeah, the `display_name` seems to only be used in the state params as far as I can tell.

acoffman: Yeah, I concur, I think that's fine for now. In general for v2 we probably want to rethink the whole name/description/display_name thing we've got going on. Its not totally clear which one to use when and where. (That's totally my fault).

PR #481, opened 2019-01-28
Attempt to retrieve diseases by name first to find instances where a user entered a disease without DOID

susannasiebert:
Closes https://github.com/griffithlab/civic-client/issues/889

### DIFF
--- a/lib/importer/disease_ontology_mirror.rb
+++ b/lib/importer/disease_ontology_mirror.rb
@@ -41,6 +41,7 @@ module Importer
                   d
                 else ( d = ::Disease.where(doid: doid).first_or_initialize)
                   d
+                end
       disease.name = name
       disease.doid = doid
       disease.display_name = display_name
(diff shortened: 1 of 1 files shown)


lib/importer/disease_ontology_mirror.rb
```diff
@@ -35,8 +35,14 @@ def create_object_from_entry(entry)
       name = entry['name']
       doid = parse_doid(entry['id'])
       synonyms = process_synonyms(entry['synonym'])
-      disease = ::Disease.where(doid: doid).first_or_initialize
+      disease = if ( d = ::Disease.find_by(doid: doid) )
+                  d
+                elsif ( d = ::Disease.find_by(name: name) )
+                  d
+                else ( d = ::Disease.where(doid: doid).first_or_initialize)
+                  d
```
acoffman: I think you're missing an `end` here.

susannasiebert: Sigh. That's what happens when you move from Python to Ruby. Fixed.

PR #482, opened 2019-01-29
Curators can reject their own submitted evidence

acoffman:
Previously only an editor could reject evidence. With this change,
curators can reject evidence too, but only evidence they themselves
submitted.

Along with https://github.com/griffithlab/civic-client/tree/feature/misc-fixes this closes griffithlab/civic-client#1037


susannasiebert:
Looks like I am 5 minutes too slow... 
We'll probably want the same behavior on assertions as well.

PR #484, opened 2019-01-30
Fix Source advanced search be PubMed ID

susannasiebert:
Fixes the 500 error mentioned in https://github.com/griffithlab/civic-client/issues/1042


susannasiebert:
@tmooney they are. But when you construct queries in plain SQL, like here, you need to use the underlying integer. This line is supposed to retrieve the correct one but you'd have to use the corrects spelling of your enum value of course.

tmooney:
Indeed it is!  What I meant by that was more in many languages such a typo could be found at compile-time.  That might be a silly notion here. 😄

PR #486, opened 2019-01-31
Initial pass at basic stats filtering.

acoffman:
These updates apply to the pie charts on the statistics page. The
endpoint now takes two optional parameters that will filter the returned
results.

limit_by_stats: can be accepted, rejected, or submitted (or ommitted)
entrez_name: entrez gene symbol (or omitted)

Supplying either or both of these parameters will limit the evidence
items from which the stats are generated.

closes griffithlab/civic-client#999 for now. May expand the
functionality later.


susannasiebert:
Just thinking about how we could amend this to allow for easier addition of new filters. If you pass the whole `params` hash to `app/models/overview_dashboard.rb` you could have two lookup tables there, one for the join table(s) for each parameter and one for the query clause and then just iterate over params to create the where and join clauses. That way you'd only need to add things to the hashes to add new filter parameters. I'm not sure what all gets stored in params so we might need to nest the query parameters under one parent key.

acoffman:
Yeah, I think I like that idea a little better - lemme see what I can come up with!

susannasiebert:
@acoffman have you had any time to refactor this? Otherwise, maybe at this point we should just merge it for the sake of getting this feature deployed and worry about refactoring if we ever expand on the existing functionality.

acoffman:
@susannasiebert Yeah, I haven't gotten a chance to refactor just yet - lets merge as is for now and I'll revisit it when we need to add more filters.

PR #488, opened 2019-02-11
Upload linkage to Allele Registry for CIViC variants that have an allele registry ID

susannasiebert:
TODO:
- Update allele registry URL to point to live site
- Add credentials to secrets.yml

### DIFF
--- a/app/jobs/update_allele_registry_ids.rb
+++ b/app/jobs/update_allele_registry_ids.rb
@@ -18,7 +18,7 @@ class UpdateAlleleRegistryIds < ActiveJob::Base
   end
 
   def add_allele_registry_link(allele_registry_id)
-    system("#{ruby_executable} #{script_path} put \"reg.test.genome.network/allele/#{allele_registry_id}/externalSource/civic\" \"p1=#{allele_registry_id}\" #{allele_registry_username} #{allele_registry_password}")
+    system("#{ruby_executable} #{script_path} put \"reg.genome.network/allele/#{allele_registry_id}/externalSource/civic\" \"p1=#{allele_registry_id}\" #{allele_registry_username} #{allele_registry_password}")
   end
 
   def ruby_executable
--- a/app/jobs/update_allele_registry_ids.rb
+++ b/app/jobs/update_allele_registry_ids.rb
@@ -2,12 +2,25 @@ class UpdateAlleleRegistryIds < ActiveJob::Base
   def perform(recurring = true)
     begin
       for v in Variant.where(allele_registry_id: nil) do
-        v.allele_registry_id = get_allele_registry_id(v)
+        allele_registry_id = get_allele_registry_id(v)
+        v.allele_registry_id = allele_registry_id
         v.save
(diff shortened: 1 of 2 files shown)


app/jobs/update_allele_registry_ids.rb
```diff
@@ -2,14 +2,27 @@ class UpdateAlleleRegistryIds < ActiveJob::Base
...
+        allele_registry_id = get_allele_registry_id(v)
+        add_allele_registry_link(allele_registry_id)
+      end
     ensure
       reschedule if recurring
     end
   end
 
+  def add_allele_registry_link(allele_registry_id)
+    system("ruby #{script_path} put \"reg.test.genome.network/allele/#{allele_registry_id}/externalSource/civic\" \"p1=#{allele_registry_id}\" #{allele_registry_username} #{allele_registry_password}")
```
acoffman: Is there a reason we're shelling out here instead of just calling this script directly? We could toss it in `lib` and just treat it like any other class in the app. Unless we're running it via CLI other places?

misc_scripts/add_allele_registry_link.rb
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env ruby
+
+require "uri"
+require 'digest/sha1'
+
+
+# create URL with authentication
+# params = vector of param=value
+def createUrl2(hostPath, params, login, password)
```
acoffman: For both createUrl methods, you could potentially use URI's builder method and ActiveSupport's to_query method to handle all of this automatically. See

https://stackoverflow.com/a/33257393
and
https://ruby-doc.org/stdlib-1.9.3/libdoc/uri/rdoc/URI/HTTP.html#method-c-build-label-Description

`to_query` should handle the escaping as needed for you, but will only be available if you move this to `lib` so `ActiveSupport` is in the library path.

Up to you though - if this works, it works and rewriting it might not be worth it.

misc_scripts/add_allele_registry_link.rb
```diff
@@ -0,0 +1,99 @@
...
+    urlCoded = "http://" + hostPathCoded + urlSuffix
+    url      = "http://" + hostPath      + urlSuffix
+    identity = Digest::SHA1.hexdigest("#{login}#{password}")
+    gbTime = Time.now.to_i.to_s
+    token = Digest::SHA1.hexdigest("#{urlCoded}#{identity}#{gbTime}")
+    request = "#{url}&gbLogin=#{URI::escape(login)}&gbTime=#{gbTime}&gbToken=#{URI::escape(token)}"
+    return request
+end
+
+def createUrl(hostPath, params)
```
acoffman: Again a minor nitpick, but generally in ruby method names are `snake_case` rather than `camelCase`.

susannasiebert:
secrets.yml on production has been updated. https://github.com/griffithlab/rails-deployments/blob/master/civic/secrets.yml as been updated. URL has been updated to point to the production allele registry.

PR #489, opened 2019-02-12
Rename "Other" Functional clinical significance term to "Unknown"

susannasiebert:
Closes https://github.com/griffithlab/civic-client/issues/928

### DIFF
--- a/app/const/constants.rb
+++ b/app/const/constants.rb
@@ -38,7 +38,7 @@ module Constants
     'Loss of Function',
     'Unaltered Function',
     'Neomorphic',
-    'Other',
+    'Unknown',
   ]
 
   DRUG_INTERACTION_TYPES = ['Combination', 'Sequential', 'Substitutes']


app/const/constants.rb
```diff
@@ -38,7 +38,7 @@ module Constants
     'Loss of Function',
     'Unaltered Function',
     'Neomorphic',
-    'Other',
+    'Unknown',
```
tmooney: Out of curiosity, I see that `:Positive` and `:Negative` in this list are symbols instead of strings.   Are those two special, or could `Neomorphic` and (now) `Unknown` also be made into symbols?  (`:Supports` in the evidence direction is also one.)

susannasiebert: Good question. I'm a bit fuzzy on when to use symbols vs strings so maybe @acoffman can shed some light on this.

acoffman: Yeah, if its one word it can be a symbol, though the savings in newer versions of ruby are much less pronounced as string literals are frozen constants anyways, so you're not doing a new object allocation all the time.

PR #491, opened 2019-02-15
Fix bug that allowed duplicate sources to be created for a source suggestion

susannasiebert:
The query as it was previously written would always return `nil` because it would search with the source_type string instead of the underlying integer. This would result in a new source to always get created when making a source suggestion, even if that source already existed in our database.

### DIFF
--- a/app/admin/utilities.rb
+++ b/app/admin/utilities.rb
@@ -150,12 +150,19 @@ ActiveAdmin.register_page 'Utilities' do
                "No Source Type provided."
              elsif (source = Source.find_by(citation_id: proposed_citation_id, source_type: proposed_source_type_int))
                "Source '#{source.description}' already present."
-             elsif proposed_source_type == 'pubmed'
+             elsif proposed_source_type == 'PubMed'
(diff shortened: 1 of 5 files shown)


susannasiebert:
@tmooney Good catch. I found a couple of more places where this occurs and fixed them.

app/admin/utilities.rb
```diff
@@ -143,19 +143,27 @@
...
+    proposed_source_type_int = Source.source_types[proposed_source_type]
     notice = if proposed_citation_id.blank?
                "No Source Id provided."
              elsif proposed_source_type.blank?
                "No Source Type provided."
-             elsif (source = Source.find_by(citation_id: proposed_citation_id, source_type: proposed_source_type))
+             elsif (source = Source.find_by(citation_id: proposed_citation_id, source_type: proposed_source_type_int))
                "Source '#{source.description}' already present."
-             elsif proposed_source_type == 'pubmed'
+             elsif proposed_source_type == 'PubMed'
```
tmooney: Bonus fix!

susannasiebert:
@acoffman almost certainly :/

acoffman:
@susannasiebert cool, i'll make an issue to clean them up when we deploy.

PR #492, opened 2019-02-19
 adds a parent attribute to org index presenter 

jmcmichael:


susannasiebert:
Required for https://github.com/griffithlab/civic-client/pull/1061 to implement https://github.com/griffithlab/civic-client/issues/1027

PR #497, opened 2019-02-25
Add additional stats

susannasiebert:
This PR adds stats about number of evidence items with/without pending changes.
It also adds some stats to break down some numbers by organization:
- organization_user_counts
- organization_badge_counts
- organization_activity_counts

Prerequisite for https://github.com/griffithlab/civic-client/issues/1077 and https://github.com/griffithlab/civic-client/issues/1071


app/models/overview_dashboard.rb
```diff
@@ -104,4 +108,38 @@ def count_by_publication_year
...
+
+  def organization_badge_count
+    BadgeAward.joins(:badge, user: [:organization])
+      .group("badge_awards.tier, badges.name, organizations.name")
+      .select("count(badge_awards.id) as badge_count, badge_awards.tier, organizations.name as org_name, badges.name as badge_name")
+      .group_by{|award| award.org_name}
+      .map{|org, awards| [org, awards.group_by{|award| award.badge_name}.map{|badge_name, awards| [badge_name, awards.map{|award| [award.tier, award.badge_count]}.to_h]}.to_h]}.to_h
+  end
+
+  def organization_activity_count
```
acoffman: The other three queries seem performant enough as-is; this last one however takes about 90 seconds to run and runs several thousand queries which explains the slowdown Josh is seeing.

There are a few things going on here that we can improve by looking at the defined relationships, looking at this line is a good place to start and we can apply the same improvements basically across the board:

```
 EvidenceItem.all.select{|e| e.submitter.organization == organization}.group_by{|e| e.status}.map{|status, es| [status, es.count]}.to_h
```

Right off the bat `EvidenceItem.all` runs a `select * from evidence_items;` and loads every single one into memory as a ruby object. Now we (essentially) have a list of in memory evidence items. 

The `.select` in this case is then not a database select but just ruby's `Enumerable#select`.  That will iterate over each evidence item object and run the block. In this case the block calls `.submitter.organization` which will run two SQL queries (one to get the submitter and then another to get the submitter's organization) for every single evidence item. We could save one query per iteration by changing the comparison to `e.submitter.organization_id == organization.id` as there's no need to compare the objects themselves, but in general, we can avoid the n+1 query entirely.

Finally, the `.group_by` in this case is again `Enumerable#group_by` rather than a SQL group. Its not the slow part of this whole thing but there's no reason we can't make the database do the heavy lifting while we're at it, Postgres is a lot faster than Ruby.

Here's an alternate implementation that fixes some of those issues:
```
  def organization_activity_count
    organizations_with_members = Organization.joins(:users).group('organizations.id, organizations.name')
      .select('organizations.name, array_agg(users.id) as member_ids')

    organizations_with_members.each_with_object({}) do |org, h|
      h[org.name] = {
        evidence_counts: EvidenceItem.joins(:submitter).where('users.id' =>  org.member_ids).group(:status).count,
        suggested_change_counts: SuggestedChange.where(user_id: org.member_ids).group(:status).count,
        assertion_count: Assertion.joins(:submitter).where('users.id' => org.member_ids).group(:status).count
      }
    end
  end
```

`organizations_with_members = Organization.joins(:users).group('organizations.id, organizations.name').select('organizations.name, array_agg(users.id) as member_ids')`

Performs a single query that gets us a list of objects that have the organization name and the list of user_ids that are members of that organization. `joins` defaults to a sql `INNER JOIN` so there will be no `nils` to worry about in this case.

`EvidenceItem.joins(:submitter).where('users.id' =>  org.member_ids).group(:status).count`

Runs a single query that joins evidence items to their submitter, limits it to only submitters in that organization, and then does a SQL group by and count, so under the hood you end up with something that looks like:

```
SELECT COUNT(*) AS count_all, "evidence_items"."status" AS evidence_items_status FROM "evidence_items" INNER JOIN "events" ON "events"."subject_id" = "evidence_items"."id" AND "events"."action" = $1 AND "events"."subject_type" = $2 INNER JOIN "users" ON "users"."id" = "events"."originating_user_id" WHERE "users"."id" IN (1, 2, 3) GROUP BY "evidence_items"."status"  [["action", "submitted"], ["subject_type", "EvidenceItem"]]
```

That produces a hash essentially identical to the one you were producing manually with the `group_by.map.to_h` chain earlier. One random thing to note is that even though the join is on "submitter" the where is on "users.id". That's a bit of an oddity because the relation defined in the model ('submitter') doesn't match the actual name of the table (just the generic users table) so that's why you see both used there. You can't join to `users` directly as it goes through the submission event.

PR #501, opened 2019-03-13
Add allele registry ID to the variant TSV dump

susannasiebert:
Closes https://github.com/griffithlab/civic-client/issues/1086

### DIFF
--- a/app/presenters/variant_tsv_presenter.rb
+++ b/app/presenters/variant_tsv_presenter.rb
@@ -31,6 +31,7 @@ class VariantTsvPresenter
       'last_review_date',
       'civic_actionability_score',
       'allele_registry_id',
+      'clinvar_ids',
     ]
   end
 
@@ -60,6 +61,7 @@ class VariantTsvPresenter
       variant.updated_at,
       variant.civic_actionability_score,
       variant.allele_registry_id,
+      variant.clinvar_entries.map(&:clinvar_id).join(','),
     ]
   end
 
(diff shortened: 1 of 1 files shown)


app/presenters/variant_tsv_presenter.rb
```diff
@@ -29,7 +29,8 @@ def self.headers
       'variant_types',
       'hgvs_expressions',
       'last_review_date',
-      'civic_actionability_score'
+      'civic_actionability_score',
+      'allele_registry_id',
```
acoffman: Generally you wouldn't want trailing commas but it doesn't really hurt anything either.

app/presenters/variant_tsv_presenter.rb
```diff
@@ -57,7 +58,8 @@ def self.row_from_object(variant)
       variant.variant_types.map(&:name).join(','),
       variant.hgvs_expressions.map(&:expression).join(','),
       variant.updated_at,
-      variant.civic_actionability_score
+      variant.civic_actionability_score,
+      variant.allele_registry_id,
```
acoffman: Same as above

app/presenters/variant_tsv_presenter.rb
susannasiebert: It's a stylistic choice but I prefer to add them so next time someone appends a field to the list, the diff only shows the new field and not the comma addition to the previous line.

PR #502, opened 2019-03-15
Add evidence item count to source presenter

susannasiebert:
Required for https://github.com/griffithlab/civic-client/issues/1045

This also fixes an issue where deleted evidence items would be included in advanced search results when we would sql-join on evidence items.


app/presenters/source_presenter.rb
```diff
@@ -22,6 +22,7 @@ def as_json(opts = {})
       status: source.status,
       is_review: source.is_review?,
       clinical_trials: source.clinical_trials.map { |t| ClinicalTrialPresenter.new(t) },
+      evidence_item_count: source.evidence_items.size,
```
acoffman: This adds a `select count(*) from evidence items....` to each source presenter we create, so anywhere we have a list of them we'll get an `n+1` query issue. That may be rare enough we don't care, but its something to be aware of.

susannasiebert: Since this is only needed for the advanced search results, we could make a new presenter that includes this, leaving the original SourcePresenter unchanged.

PR #508, opened 2019-04-26
Add rate limiting using rack-attack gem

acoffman:


susannasiebert:
We should also figure out what needs to be changed to make the tests pass again.

acoffman:
Hah, yeah, was just working on the tests passing - it doesn't attach the throttling data when the requests come from the test framework, but that should be fixed now. The limit is semi arbitrary, I mostly looked at what the defaults were for this framework, what other major APIs set (Github is ~1.4/second) to make sure we're within what's considered reasonable and then looked at what our server was doing right now. We can definitely adjust as needed and we can whitelist or set different limits for different IPs if needed.

PR #511, opened 2019-05-17
Refactor the UpdateAlleleRegistryIds job and add new SetAlleleRegistryIds

susannasiebert:
The UpdateAlleleRegistryIDs job now correctly updates the allele registry ID on a variant where it changed (e.g. if the coordinates were updated) and only makes the post request to the allele registry API if the ID changed. This occurs weekly to reduce the number of requests to their API. We might be able to get away with running it daily if we don't expect this to happen very often.

This PR also add a SetAlleleRegistryIds job which runs daily to initially set the allele registry ID for new variants.

I'm in contact with Ronak on how to remove reverse linkouts for allele registry IDs we no longer match (e.g. after a variant's coordinates were updated)

### DIFF
--- a/app/jobs/set_allele_registry_ids.rb
+++ b/app/jobs/set_allele_registry_ids.rb
@@ -1,0 +1,25 @@
+require 'rbconfig'
+
+class SetAlleleRegistryIds < AlleleRegistryIds
+  def perform(recurring = true)
+    begin
+      for v in Variant.where(allele_registry_id: nil) do
+        allele_registry_id = get_allele_registry_id(v)
+        v.allele_registry_id = allele_registry_id
+        v.save
+        add_allele_registry_link(allele_registry_id)
+      end
+    ensure
+      reschedule if recurring
+    end
+  end
+
+  def reschedule
+    self.class.set(wait_until: next_day).perform_later
+  end
+
+  def next_day
+    Date.tomorrow
+      .midnight
+  end
+end
--- a/app/jobs/set_allele_registry_ids.rb
+++ b/app/jobs/set_allele_registry_ids.rb
@@ -3,7 +3,7 @@ require 'rbconfig'
 class SetAlleleRegistryIds < AlleleRegistryIds
   def perform(recurring = true)
     begin
-      for v in Variant.where(allele_registry_id: nil) do
+      Variant.where(allele_registry_id: nil).each do |v|
         allele_registry_id = get_allele_registry_id(v)
         v.allele_registry_id = allele_registry_id
         v.save
--- a/app/jobs/update_allele_registry_ids.rb
+++ b/app/jobs/update_allele_registry_ids.rb
@@ -11,7 +11,7 @@ class UpdateAlleleRegistryIds < AlleleRegistryIds
           v.save
           add_allele_registry_link(allele_registry_id)
           #delete the linkout if no other variant has this allele registry ID
-          if Variant.find_by(allele_registry_id: old_allele_registry_id).exists?
+          if Variant.where(allele_registry_id: old_allele_registry_id).exists?
               delete_allele_registry_link(old_allele_registry_id)
           end
         end
--- a/app/jobs/update_allele_registry_ids.rb
+++ b/app/jobs/update_allele_registry_ids.rb
@@ -3,7 +3,7 @@ require 'rbconfig'
 class UpdateAlleleRegistryIds < AlleleRegistryIds
   def perform(recurring = true)
     begin
-      for v in Variant.where.not(allele_registry_id: nil) do
+      Variant.where.not(allele_registry_id: nil).each do |v|
         old_allele_registry_id = v.allele_registry_id
         allele_registry_id = get_allele_registry_id(v)
         if allele_registry_id != old_allele_registry_id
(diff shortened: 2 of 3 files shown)


susannasiebert:
Ronak replied to `just send a DELETE request on the path : reg.clinicalgenome.org/allele/{CAid}/externalSource/civic`

app/jobs/set_allele_registry_ids.rb
```diff
@@ -0,0 +1,25 @@
+require 'rbconfig'
+
+class SetAlleleRegistryIds < AlleleRegistryIds
+  def perform(recurring = true)
+    begin
+      for v in Variant.where(allele_registry_id: nil) do
```
acoffman: I didn't actually realize you could do this! `.each` would certainly be more idiomatic, but if this works, I don't have a problem with it.

app/jobs/update_allele_registry_ids.rb
```diff
@@ -1,72 +1,34 @@
...
       for v in Variant.where.not(allele_registry_id: nil) do
+        old_allele_registry_id = v.allele_registry_id
         allele_registry_id = get_allele_registry_id(v)
-        add_allele_registry_link(allele_registry_id)
+        if allele_registry_id != old_allele_registry_id
+          v.allele_registry_id = allele_registry_id
+          v.save
+          add_allele_registry_link(allele_registry_id)
+          #delete the linkout if no other variant has this allele registry ID
+          if Variant.find_by(allele_registry_id: old_allele_registry_id).blank?
```
acoffman: You could use `exists?` here instead of `blank?` which will do a `SQL` exists query instead of actually loading an object if it finds one.

app/jobs/set_allele_registry_ids.rb
susannasiebert: It works but it is actually not recommended. I think I was a bit out in Python-land when I wrote it. I will change it to the more idiomatic version.

susannasiebert:
```suggestion
      Variant.where(allele_registry_id: nil).each do |v|
```

app/jobs/update_allele_registry_ids.rb
```diff
@@ -1,72 +1,34 @@
...
-class UpdateAlleleRegistryIds < ActiveJob::Base
+class UpdateAlleleRegistryIds < AlleleRegistryIds
   def perform(recurring = true)
     begin
-      for v in Variant.where(allele_registry_id: nil) do
-        allele_registry_id = get_allele_registry_id(v)
-        v.allele_registry_id = allele_registry_id
-        v.save
-      end
       for v in Variant.where.not(allele_registry_id: nil) do
```
susannasiebert:
```suggestion
      Variant.where.not(allele_registry_id: nil).each do |v|
```

app/jobs/update_allele_registry_ids.rb
```diff
@@ -1,72 +1,34 @@
...
-      end
-      for v in Variant.where.not(allele_registry_id: nil) do
-        allele_registry_id = get_allele_registry_id(v)
-        add_allele_registry_link(allele_registry_id)
+        if allele_registry_id != old_allele_registry_id
+          v.allele_registry_id = allele_registry_id
+          v.save
+          add_allele_registry_link(allele_registry_id)
+          #delete the linkout if no other variant has this allele registry ID
+          if Variant.find_by(allele_registry_id: old_allele_registry_id).blank?
```
susannasiebert:
```suggestion
          if Variant.find_by(allele_registry_id: old_allele_registry_id).exists?
```

acoffman: This one is my bad for not explaining it better, sorry to create extra work!

`find_by` eagerly resolves; ie once you call it, you either have `nil` or your `Variant` instance, its not lazy like most ActiveRecord query methods, so if you stick with `find_by` then, yeah, you'll want to use `blank?` because you no longer have a query object, you just have your result. 

`where` on the other hand is lazy like the rest of the query methods, so calling `Variant.where(foo: bar)` returns an ActiveRecord query object that you can keep building on, and it only actually runs the query and resolves when you force it to (by iterating over the results, or calling `.all` or whatever). So in order to use `exists?` you'll want to switch to `.where` or you can just revert it all and its not a big deal either.

app/jobs/update_allele_registry_ids.rb
```diff
@@ -1,72 +1,34 @@
...
-      end
-      for v in Variant.where.not(allele_registry_id: nil) do
-        allele_registry_id = get_allele_registry_id(v)
-        add_allele_registry_link(allele_registry_id)
+        if allele_registry_id != old_allele_registry_id
+          v.allele_registry_id = allele_registry_id
+          v.save
+          add_allele_registry_link(allele_registry_id)
+          #delete the linkout if no other variant has this allele registry ID
+          if Variant.find_by(allele_registry_id: old_allele_registry_id).exists?
```
susannasiebert:
```suggestion
          if Variant.where(allele_registry_id: old_allele_registry_id).exists?
```

app/jobs/update_allele_registry_ids.rb
susannasiebert: Thanks for the explanation. This seems to be a thing I do a lot (loading the instances instead of working with the ActiveRecord objects). I will try to be more mindful of this in the future.

PR #513, opened 2019-06-21
Support gene names in the direct linking

susannasiebert:
Closes https://github.com/griffithlab/civic-client/issues/1118

This adds support for direct linking to genes by their entrez gene name using the civicdb.org/links/entrez_name/:name URL format. Special characters of the gene name (like `/`) will need to be URL-encoded.

### DIFF
--- a/app/models/frontend_router.rb
+++ b/app/models/frontend_router.rb
@@ -28,8 +28,10 @@ class FrontendRouter
       [ Variant, :id, ]
     when /evidence/, /evidence_items?/
       [ EvidenceItem, :id, ]
-    when /entrez/
+    when /entrez_id/
       [ Gene, :entrez_id, ]
+    when /entrez_name/
+      [ Gene, :name , ]
     when /variant_groups?/
       [ VariantGroup, :id, ]
     when /revisions?/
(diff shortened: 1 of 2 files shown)


app/models/frontend_router.rb
```diff
@@ -28,8 +28,10 @@ def query_info
       [ Variant, :id, ]
     when /evidence/, /evidence_items?/
       [ EvidenceItem, :id, ]
-    when /entrez/
```
acoffman: Even though its not totally clear, I think we should leave this `entrez` unfortunately as we don't want to break any links that currently work. We'd need to change the regex to be `entrez$` or similar so it doesn't match `entrez_name` accidentally.

The other option is that we could just keep `entrez` and then allow them to pass an ID or Symbol and do the resolution on our end.

susannasiebert: The API docs says that the "official" endpoint is `entrez_id` (https://griffithlab.github.io/civic-api-docs/#linking). With that in mind, I think we should actually make the behavior match the documentation and fix any broken links. If we want to support `entrez` instead of `entrez_id` then we should update the documentation to reflect that.

acoffman: Good catch, agreed! I didn't realize that's what the docs already said. With that in mind, I think this is good then.

PR #514, opened 2019-06-27
New job for weekly updating clinical trial ids from PubMed

susannasiebert:
Closes https://github.com/griffithlab/civic-client/issues/1116

### DIFF
--- a/app/jobs/update_clinical_trials.rb
+++ b/app/jobs/update_clinical_trials.rb
@@ -1,0 +1,28 @@
+class UpdateClinicalTrials < ActiveJob::Base
+  def perform(recurring = true)
+    begin
+      Source.where(source_type: 'PubMed').each do |source|
+        resp = Scrapers::PubMed.call_pubmed_api(source.citation_id)
+        clinical_trials = resp.clinical_trial_ids.uniq.map do |nct_id|
+          ClinicalTrial.where(nct_id: nct_id).first_or_create
+        end
+        source.clinical_trials = clinical_trials
+        source.save
+        sleep 0.1
+      end
+    ensure
+      reschedule if recurring
+    end
+  end
+
+  def reschedule
+    self.class.set(wait_until: next_week).perform_later
+  end
+
+  def next_week
+    Date.today
+      .beginning_of_week
+      .next_week
+      .midnight
+  end
+end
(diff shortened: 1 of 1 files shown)


app/jobs/update_clinical_trials.rb
```diff
@@ -0,0 +1,28 @@
...
+  def perform(recurring = true)
+    begin
+      Source.where(source_type: 'PubMed').each do |source|
+        resp = Scrapers::PubMed.call_pubmed_api(source.citation_id)
+        clinical_trials = resp.clinical_trial_ids.uniq.map do |nct_id|
+          ClinicalTrial.where(nct_id: nct_id).first_or_create
+        end
+        source.clinical_trials = clinical_trials
+        source.save
+        sleep 0.1
```
acoffman: We may want a slightly longer sleep here unfortunately. On the plus side, since it happens in the background the slowness won't really matter.

app/jobs/update_clinical_trials.rb
```diff
@@ -0,0 +1,28 @@
+class UpdateClinicalTrials < ActiveJob::Base
+  def perform(recurring = true)
+    begin
+      Source.where(source_type: 'PubMed').each do |source|
```
acoffman: I wonder if these ever change after they've been added or if it would be safe to just grab Pubmed sources that don't have trials associated with them.

susannasiebert: Not so much change but I could imagine additional ones being added at a later date.

app/jobs/update_clinical_trials.rb
susannasiebert: How long do you think would be best?

acoffman: Hrm, don't have a strong opinion. Maybe a quarter second or a half second?

app/jobs/update_clinical_trials.rb
acoffman: Yeah good point, I'd just leave it as is.

app/jobs/update_clinical_trials.rb
susannasiebert:
```suggestion
        sleep 0.5
```

PR #521, opened 2019-08-01
Remove rescheduling from ensure blocks.

acoffman:
This will prevent runaway job failures like we saw with the sequence ontology update task. `after_perform` is only called if the job succeeds, so we will no longer retry in `delayed_job` AND reschedule a duplicate copy. Also reduce the number of retries from the default 25 to a more
sane 3.

Really, we should probably extract a "recurring job" mixin out of these, as the copy-paste job becomes pretty obvious here. There are some non-trivial changes to the `ActiveJob` API in newer versions of rails however, so I will defer that task to when we upgrade anyways. Logged an issue for the task here: griffithlab/civic-client#1152


app/jobs/fix_pubmed_entries_missing_authors.rb
```diff
@@ -1,11 +1,19 @@
 class FixPubmedEntriesMissingAuthors < ActiveJob::Base
+
+  attr_reader :recurring
+
+  after_perform do |job|
```
susannasiebert: Should we make a parent class that pulls out the after_perfom?

susannasiebert: (just re-read your pull request description and I agree that we should add a mixin or parent class to pull out the after_perfom)

PR #526, opened 2019-08-27
Add job to create the civicpy pkl file

susannasiebert:
This requires https://github.com/griffithlab/civicpy/pull/26


susannasiebert:
griffithlab/civicpy#26 hasn't been merged yet so this will fail. We'd also need a new CIViCpy release.

PR #529, opened 2019-08-29
Limit drug name suggestions

acoffman:
Only return local name suggestions in the typeahead for drugs already
associated with an accepted evidence item.

### DIFF
--- a/app/models/drug_name_suggestion.rb
+++ b/app/models/drug_name_suggestion.rb
@@ -17,8 +17,11 @@ class DrugNameSuggestion
   end
 
   def self.get_local_suggestions(name)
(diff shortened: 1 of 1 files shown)


susannasiebert:
I know this has already been merged but I think this is a bad idea once we start scraping NCIT. The whole point of doing that is to have suggestions available for things that aren't already used so that we don't get multiple different versions of the same drug with slightly different (miss)spellings. If we are concerned about the leftover bad drugs that remain after a suggested change is accepted, a better solution would be to have a job that cleans those up regularly, i.e. delete all drugs without an NCIT id and no associated evidence items.

PR #535, opened 2019-09-17
WIP: Add script to write CIViC VCF

susannasiebert:
TODO:
- [x] Add an app/jobs to create the VCF.


acoffman:
We'll want to remember to install civicpy on the prod server when we deploy this.

susannasiebert:
I need to check how we did it for Ndex but I vaguely remember that we modified the Capistrano deploy task to include python package installations.

susannasiebert:
The civicpy dependency has already been added [here](lib/capistrano/tasks/dependencies.rake)

PR #541, opened 2019-10-08
Updates to VCF and pkl file creation jobs

susannasiebert:
This PR updates the jobs to create a CIViC VCF and CIViCpy cache pkl to 
- Create monthly and nightly versions
- Create an accepted-only VCF and accepted-and-submitted VCF every time it runs


lib/capistrano/tasks/dependencies.rake
```diff
@@ -4,7 +4,7 @@ namespace :dependencies do
     on roles(:all) do
       execute "sudo apt-get -y install python python3-pip"
       execute :pip3, "install ndexutil -r #{current_path}/misc_scripts/ndex_requirements.txt --no-cache-dir --no-deps"
-      execute :pip3, "install civicpy"
+      execute :pip3, "install civicpy==1.0.0rc1"
```
ahwagner: we should bump this to 1.0.0rc2, where we can implement the new cache download path. At this point we cannot re-release rc1, as we have publicized CIViCpy.

susannasiebert:
```suggestion
      execute :pip3, "install civicpy==1.0.0rc2"
```

susannasiebert: Done.

PR #550, opened 2019-11-05
Pull in MANE preferred transcript from the allele registry

susannasiebert:


app/models/mane_preferred_transcript.rb
```diff
@@ -0,0 +1,51 @@
...
+  def get_mane_preferred_ref_seq(variant)
+    gene_response = make_gene_request(variant.gene.name)
+    JSON.parse(gene_response)[0]['externalRecords']['MANEPrefRefSeq']['id'] rescue nil
+  end
+
+  private
+  def make_gene_request(gene_name)
+    Scrapers::Util.make_get_request(gene_url(gene_name))
+  rescue StandardError
+    {}
```
acoffman: I believe this will break when the return value is passed to `JSON.parse` which expects a string. I would either return a string `{}` or do the parsing in this method and catch either error here.

app/models/mane_preferred_transcript.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def make_allele_request(allele_registry_id)
+    Scrapers::Util.make_get_request(allele_url(allele_registry_id))
+  rescue StandardError
```
acoffman: Same as above

app/models/mane_preferred_transcript.rb
susannasiebert: I just pushed a commit to fix this.

PR #554, opened 2019-12-06
Make EvidenceItem and Assertion responses to changes consistent

acoffman:
Now, if an evidence item or assertion is rejected and someone proposes a
change to it, it will return to 'submitted' status. Likewise accepting a
change to a rejected item will return it to 'submitted' status.

closes griffithlab/civic-client#1256

### DIFF
--- a/app/models/concerns/moderated.rb
+++ b/app/models/concerns/moderated.rb
@@ -33,6 +33,7 @@ module Moderated
       raise NoSuggestedChangesError
     else
       SuggestedChange.create(user: user, suggested_changes: c, moderated: self).tap do
+        on_change_suggested
         reload
       end
     end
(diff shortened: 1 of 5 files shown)


app/models/concerns/moderated.rb
```diff
@@ -117,6 +118,9 @@ def validate_additional_changeset(changes)
   def after_change_accept(change)
   end
 
+  def on_change_suggested
```
susannasiebert: Should this throw an error if it's not overwritten in the class that uses this mixin?

acoffman: Not in this case, as it will get called every time, but if the implementing class doesn't want to use the callback then a no-op is just fine.

PR #555, opened 2019-12-18
Catch errors in CIViCpy jobs before rescheduling

susannasiebert:
The CIViCpy-related jobs are rescheduling and not marked as failed but don't appear to actually create the files. Hopefully, this will catch this and allows us to debug the reason for them not working right.

### DIFF
--- a/app/jobs/create_civic_vcfs.rb
+++ b/app/jobs/create_civic_vcfs.rb
@@ -6,7 +6,7 @@ class CreateCivicVcfs < ActiveJob::Base
   end
 
   def perform(recurring = true)
-    @recurring if recurring
+    @recurring = recurring
     execute
   end
 
--- a/app/jobs/create_civic_vcfs.rb
+++ b/app/jobs/create_civic_vcfs.rb
@@ -1,7 +1,13 @@
 class CreateCivicVcfs < ActiveJob::Base
-  def perform
+  attr_reader :recurring
+
+  after_perform do |job|
+    job.reschedule if job.recurring
+  end
+
+  def perform(recurring = true)
+    @recurring if recurring
     execute
-    reschedule
   end
 
   def execute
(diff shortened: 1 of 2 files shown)


app/jobs/create_civic_vcfs.rb
```diff
@@ -1,7 +1,13 @@
 class CreateCivicVcfs < ActiveJob::Base
-  def perform
+  attr_reader :recurring
+
+  after_perform do |job|
+    job.reschedule if job.recurring
+  end
+
+  def perform(recurring = true)
+    @recurring if recurring
```
acoffman: You'll want to assign `@recurring` the value of `recurring` here, this line doesn't do anything as written.

app/jobs/create_civicpy_cache_pkl.rb
```diff
@@ -1,7 +1,13 @@
 class CreateCivicpyCachePkl < ActiveJob::Base
-  def perform
+  attr_reader :recurring
+
+  after_perform do |job|
+    job.reschedule if job.recurring
+  end
+
+  def perform(recurring = true)
+    @recurring if recurring
```
acoffman: Same here - looks good otherwise though. Good to merge after fixing!

app/jobs/create_civic_vcfs.rb
susannasiebert:
```suggestion
    @recurring = recurring
```

app/jobs/create_civicpy_cache_pkl.rb
susannasiebert:
```suggestion
    @recurring = recurring
```

app/jobs/create_civic_vcfs.rb
susannasiebert: Duh. I fixed it now.

PR #556, opened 2019-12-18
Use GeneDetailPresenter for gene advanced search to include lifecycle actions

susannasiebert:
This is needed for CIViCpy (https://github.com/griffithlab/civicpy/issues/65).


susannasiebert:
I fixed the eager load and created a new presenter as suggested.

PR #564, opened 2020-01-03
Switch from delayed_job to sidekiq

acoffman:
Use the sidekiq cron gem for scheduled tasks instead of the less reliable self-rescheduling tasks. Sidekiq's author recommends using systemd or similar rather than `cap` to handle restarting the workers, so we'll need to set that up or add our own `cap` tasks. 

This will also provide a web ui to view scheduled tasks, failures, etc.


acoffman:
Screen Shot 2020-01-09 at 10 16 48 AM
Screen Shot 2020-01-09 at 10 16 55 AM
Screen Shot 2020-01-09 at 10 17 11 AM


Screenshots of new UI for managing queues and scheduled tasks.

PR #566, opened 2020-01-07
If no DOID is present, provide no link.

acoffman:
This is as opposed to the previous link that was just to the homepage.

Supports griffithlab/civic-client#1271

### DIFF
--- a/app/presenters/disease_presenter.rb
+++ b/app/presenters/disease_presenter.rb
@@ -19,7 +19,7 @@ class DiseasePresenter
     if disease.doid.present?
       "http://www.disease-ontology.org/?id=DOID:#{disease.doid}"
     else
-      "http://www.disease-ontology.org/"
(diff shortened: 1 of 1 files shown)


jmcmichael:
Oddly, the DO homepage is still returned even if DOID is null. I switched to using the presence of the DOID instead of the URL for the template logic, for [the client PR](griffithlab/civic-client/pull/1283) which isn't as elegant but ultimately has the same result.

susannasiebert:
@jmcmichael That's strange. I can confirm that on my end where I spot checked things, this appears to be working. Do you have an example/screenshot where it isn't working for you?

Screen Shot 2020-01-08 at 8 18 18 AM

jmcmichael:
OK it's working as expected now, I must have forgotten to restart rails yesterday evening. Thanks!

PR #568, opened 2020-01-07
Make nccn_guideline on assertions a foreign key table instead of enum

susannasiebert:
Prerequisite for https://github.com/griffithlab/civic-client/issues/1255.

This will require some frontend changes to get suggested changes to work. There will also need to be a second database migration to remove `nccn_guideline_old`. We will also probably want to add an endpoint that returns all available NCCN guidelines so that we can stop hardcoding them on the client side.

### DIFF
--- a/app/controllers/assertion_moderations_controller.rb
+++ b/app/controllers/assertion_moderations_controller.rb
@@ -14,8 +14,11 @@ class AssertionModerationsController < ModerationsController
     if params[:variant].present?
       params[:variant] = params[:variant][:name]
     end
+    if params[:nccn_guideline].present?
+      params[:nccn_guideline] = params[:nccn_guideline][:name]
+    end
 
-    params.permit(:summary, :description, :gene, :variant, :disease, :nccn_guideline, :nccn_guideline_version, :evidence_type, :evidence_direction, :amp_level, :clinical_significance, :drug_interaction_type, :fda_regulatory_approval, :fda_companion_test, :variant_origin)
+    params.permit(:summary, :description, :gene, :variant, :disease, :nccn_guideline_version, :evidence_type, :evidence_direction, :amp_level, :clinical_significance, :drug_interaction_type, :fda_regulatory_approval, :fda_companion_test, :variant_origin)
   end
 
   def additional_moderation_params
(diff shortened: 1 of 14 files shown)


app/models/nccn_guideline.rb
```diff
@@ -0,0 +1,3 @@
+class NccnGuideline < ActiveRecord::Base
+  has_many :assertions
```
acoffman: Do we need to add the corresponding relation in `Assertion`?

app/controllers/assertion_moderations_controller.rb
```diff
@@ -14,8 +14,11 @@ def moderation_params
     if params[:variant].present?
       params[:variant] = params[:variant][:name]
     end
+    if params[:nccn_guideline].present?
+      params[:nccn_guideline] = params[:nccn_guideline][:name]
+    end
 
-    params.permit(:summary, :description, :gene, :variant, :disease, :nccn_guideline, :nccn_guideline_version, :evidence_type, :evidence_direction, :amp_level, :clinical_significance, :drug_interaction_type, :fda_regulatory_approval, :fda_companion_test, :variant_origin)
+    params.permit(:summary, :description, :gene, :variant, :disease, :nccn_guideline_version, :evidence_type, :evidence_direction, :amp_level, :clinical_significance, :drug_interaction_type, :fda_regulatory_approval, :fda_companion_test, :variant_origin)
```
acoffman: I'd have to double check but with this as a new entity and not a field directly on Assertions anymore, do we need to update the suggested change logic to handle this appropriately? (more like a drug or source)

app/models/nccn_guideline.rb
susannasiebert: It's there: https://github.com/griffithlab/civic-server/blob/issue_1255/app/models/assertion.rb#L15

app/controllers/assertion_moderations_controller.rb
susannasiebert: This is not a many-to-many relationship so this should still work; it's more like diseases in that regard. (I don't think I was supposed to remove it from this list)

acoffman: Ah yeah, assuming this works as written, i think its good to go then.

PR #577, opened 2020-01-23
Replace ActiveRecord::Type::Boolean.new.type_cast_from_user with ActiveRecord::Type::Boolean.new.cast

susannasiebert:
`type_cast_from_user` was replaced with `cast` in Rails 5

Closes https://github.com/griffithlab/civic-client/issues/1299


susannasiebert:
I think the idea was to support multiple options like `true`, `1` etc and since they are submitted as strings I added this parsing.

PR #582, opened 2020-01-30
Implement a gene name exact_match option

susannasiebert:
Required backend changes for https://github.com/griffithlab/civic-client/issues/887

### DIFF
--- a/app/controllers/genes_controller.rb
+++ b/app/controllers/genes_controller.rb
@@ -120,7 +120,8 @@ class GenesController < ApplicationController
 
   def name_search(query)
     if params[:name].present?
-      if params[:exact_match].present? and ActiveRecord::Type::Boolean.new.cast(params[:exact_match])
+      binding.pry
+      if ActiveRecord::Type::Boolean.new.cast(params[:exact_match])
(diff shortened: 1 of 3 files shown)


app/controllers/genes_controller.rb
```diff
@@ -120,8 +120,12 @@ def entrez_search(query)
 
   def name_search(query)
     if params[:name].present?
-      query.where('genes.name ILIKE :name OR gene_aliases.name ILIKE :name', name: "#{params[:name]}%")
-        .reorder('char_length(genes.name) asc')
+      if params[:exact_match].present? and ActiveRecord::Type::Boolean.new.cast(params[:exact_match])
```
acoffman: I think this could be simplified if you like by removing the presence check, but that's personal preference. The `ActiveRecord::Type::Boolean` will return nil given nil which would be falsy. But if you think this is clearer, I'm fine sticking with it as well.

susannasiebert: Good point. We're repeating this pattern in a couple of places so I removed it everywhere it occurs.

PR #591, opened 2020-02-14
Update event presenter to include event organization and user role

susannasiebert:


app/presenters/event_presenter.rb
```diff
@@ -15,7 +15,9 @@ def as_json(options = {})
       description: description,
       state_params: event.state_params,
       user: user,
-      unlinkable: event.unlinkable
+      unlinkable: event.unlinkable,
+      organization: event.organization ? OrganizationIndexPresenter.new(event.organization) : nil,
```
acoffman: This line will add a `select * from organizations where organization.id = event.organization_id` to every single instance of the event presenter. We likely want to look at places where we're calling this and add an `eager_load` or `include` to the initial query to prevent that or we'll get an extra db query for every row in the table.

susannasiebert: I found a few places that uses this presenter under the hood and added `:organization` in their already existing `includes` statements.

acoffman:
This looks good - only one more thing - where we use the minimal presenter, we no longer need to load the organization associated with the user. Otherwise, merge away!

susannasiebert:
@acoffman I removed the include for the user's organization in the one place where I found it. Everywhere else (as far as I was able tot tell) we were only pulling in the originating user and not their organization.

acoffman:
Yep, you're totally right - somehow I just read it incorrectly.

PR #595, opened 2020-03-11
Fix bug that would throw an error if assertion was submitted without a NCCN guideline

susannasiebert:
Closes https://github.com/griffithlab/civic-client/issues/1337


jmcmichael:
Tested, looks good to me!

PR #598, opened 2020-04-10
Support direct links using CIViC IDs

susannasiebert:
Closes https://github.com/griffithlab/civic-client/issues/1356

### DIFF
--- a/app/models/frontend_router.rb
+++ b/app/models/frontend_router.rb
@@ -40,6 +40,23 @@ class FrontendRouter
       [ Assertion, :id ]
     when /allele_registry/
       [ Variant, :allele_registry_id, ]
+    when /id/
+      type, parsed_id = id.upcase.match(/^(AID|GID|VID|EID|SID)(\d+)$/).captures
+      id = parsed_id
+      case type
+      when "AID"
+        [ Assertion, :id ]
+      when "GID"
+        [ Gene, :id ]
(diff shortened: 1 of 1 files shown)


app/models/frontend_router.rb
```diff
@@ -40,6 +40,23 @@ def query_info
       [ Assertion, :id ]
     when /allele_registry/
       [ Variant, :allele_registry_id, ]
+    when /id/
+      type, parsed_id = id.upcase.match(/^(AID|GID|VID|EID|SID)(\d+)$/).captures
+      id = parsed_id
```
acoffman, line 45: Does this actually change the value of `@id`? I don't believe it will. You'll want to either change the instance variable directly or change `attr_reader` into an `attr_accessor` and then do `self.id = `

susannasiebert: I did test it and it appeared to work but I will do more digging before we merge this in.

acoffman: If it works I'm fine with it; I'm just not sure how it does given this:
```
class A
  attr_reader :foo

  def initialize
    @foo = 1
  end

  def bar
    foo = 2
  end
end

x = A.new
puts x.foo
x.bar()
puts x.foo
```
prints 

```
1
1
```

I must be missing some context somewhere :-D

PR #599, opened 2020-04-14
Pass along "acting as" organization context for pundit

acoffman:
This wraps the `User` object that will be passed to pundit authz checks with additional context from the request. In this case its just the organization that the user is "acting on behalf of" for a particular action but we can imagine expanding that as needed for more complex logic.

This also updates the policies for everything I'm aware of that generates an attributable event to check that either the request hasn't specified an organization to act on behalf of or, if it does, the logged in user is in fact a member.

I am definitely open to suggestions on naming because I really don't like the current one 😀

### DIFF
--- a/app/models/user_context.rb
+++ b/app/models/user_context.rb
@@ -1,0 +1,12 @@
+class UserContext < DelegateClass(User)
+  attr_reader :acting_as_organization_id
+
+  def initialize(user, organization_id)
+    @acting_as_organization_id = organization_id
+    super(user)
+  end
+
+  def acting_as_organization
+    Organization.find(acting_as_organization_id)
+  end
+end
--- a/app/models/user_context.rb
+++ b/app/models/user_context.rb
@@ -1,9 +1,9 @@
 class UserContext < DelegateClass(User)
-  attr_reader :acting_as_organization_id
+  attr_reader :action_organization
 
   def initialize(user, organization_id)
     numeric_id = Integer(organization_id) rescue nil
-    @acting_as_organization_id = numeric_id
+    @action_organization = numeric_id
     super(user)
   end
 end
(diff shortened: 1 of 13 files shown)


app/models/user_context.rb
```diff
@@ -0,0 +1,12 @@
+class UserContext < DelegateClass(User)
+  attr_reader :acting_as_organization_id
+
+  def initialize(user, organization_id)
+    @acting_as_organization_id = organization_id
+    super(user)
+  end
+
+  def acting_as_organization
+    Organization.find(acting_as_organization_id)
```
susannasiebert, line 10: Would this fail if an action is submitted without an acting organization (i.e. a user doesn't belong to an organization)?

susannasiebert:
Instead of `acting_as_organization` maybe something like `event_organization` or `action_organization` would work?

app/models/user_context.rb
acoffman: Indeed it would. Actually I don't think it needs to exist really - I added it as a convenience method and then didn't actually end up using it.

susannasiebert:
@acoffman just FYI, I just pushed a commit for a missing include of the `PolicyHelpers` in one of the policy files.

app/policies/policy_helpers.rb
```diff
@@ -9,11 +9,11 @@ def valid_coi_statement(user)
     statement.present? && !statement.expired?
   end
 
-  def belongs_to_acting_as_organization?(user)
-    if user.acting_as_organization_id.nil?
+  def belongs_to_action_organization?(user)
```
susannasiebert, line 12: should the be broken out maybe into two things `belongs_to_action_organization` and `has_no_organization`?

acoffman: Yeah, I'm not sure what the best way to handle this is, if the api request doesn't specify an organization then that's fine and the check should pass,  but you're right that its perhaps two conceptually different things

PR #607, opened 2020-05-20
Only update/set a variant's allele registry ID if API query returned one

susannasiebert:
This is related to https://github.com/griffithlab/civic-client/issues/1395 but doesn't cause any fatal errors so wouldn't be the reason for variants not getting their allele registry id set. However, this will improve runtime drastically.

### DIFF
--- a/app/jobs/set_allele_registry_ids.rb
+++ b/app/jobs/set_allele_registry_ids.rb
@@ -8,7 +8,7 @@ class SetAlleleRegistryIds < AlleleRegistryIds
       if allele_registry_id == '_:CA'
         v.allele_registry_id = 'unregistered'
         v.save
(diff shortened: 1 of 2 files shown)


app/jobs/set_allele_registry_ids.rb
```diff
@@ -8,7 +8,7 @@ def perform
       if allele_registry_id == '_:CA'
         v.allele_registry_id = 'unregistered'
         v.save
-      else
+      elsif not allele_registry_id.nil?
```
acoffman, line 11: You if you keep this a negative condition you might want to switch to `!` instead of `not`, otherwise you could swap it to `if allele_registry_id.present?`

PR #617, opened 2020-06-22
Overwrite Event organization for single-org users if no org is passed along by request

susannasiebert:
The frontend should submit the organization of a user to attribute an Event to but it won't if someone runs an outdated version. This will at least catch cases for single-org users.


app/models/actions/propose_assertion.rb
```diff
@@ -26,16 +27,16 @@ def execute
...
-        organization: organization
-      )
+      @subject = assertion
+      create_event('assertion submitted')
       assertion.subscribe_user(originating_user)
       @assertion = assertion
     end
 
+    def state_params
+      nil
```
acoffman, line 37: Do assertions not have state params? We do link to them from the events right?

app/models/actions/propose_evidence_item.rb
```diff
@@ -22,17 +23,17 @@ def execute
...
-      )
+      @subject = evidence_item
+      create_event('submitted')
       evidence_item.subscribe_user(originating_user)
       process_source_suggestion
       @evidence_item = evidence_item
     end
 
+    def state_params
+      nil
```
acoffman, line 34: Same question here

app/models/actions/suggest_publication.rb
```diff
@@ -80,13 +82,8 @@ def create_publication_submission
...
 
-    def create_event
-      Event.create(
-        action: 'publication suggested',
-        originating_user: originating_user,
-        subject: source,
-        organization: organization
-      )
+    def state_params
+      nil
```
acoffman, line 86: Ditto

app/models/actions/propose_assertion.rb
susannasiebert: The deleted portion that created the Event previously wasn't setting the state params. I'm confused what differentiates the Event's state params from the subject's state params. The subject/assertion themselves does have state params.

PR #621, opened 2020-06-26
Fix broken filters on evidence item and assertion browse tables

susannasiebert:
Closes https://github.com/griffithlab/civic-client/issues/1419

The evidence item and assertion browse table backends had some special logic baked in to handle enums like the evidence type, clinical significance, etc. However, when applying that special logic it would subsequently delete the parameter for that filter so it wouldn't try to also apply these filters again during the generic filtering logic. This would cause these filters to be missing when the total was being calculated. 

This change removes the deleting of these filter parameters from the parameter hash and instead implements an alternate approach on how to skip these type of filter parameters during the main filtering logic.

### DIFF
--- a/app/datatables/datatable_base.rb
+++ b/app/datatables/datatable_base.rb
@@ -34,9 +34,7 @@ class DatatableBase
 
   def filter(objects)
     if params['filter']
-      filter_params = params['filter'].reject{|name, value| special_filters.include? name}
-    end
-    if filter_params
+      filter_params = params['filter'].reject{|name, _| special_filters.include? name}
       or_filters = []
       or_values = []
       filtered_objects = filter_params.inject(objects) do |o, (col, term)|
--- a/app/datatables/datatable_base.rb
+++ b/app/datatables/datatable_base.rb
@@ -33,7 +33,10 @@ class DatatableBase
   end
 
   def filter(objects)
-    if filter_params = params['filter']
+    if params['filter']
+      filter_params = params['filter'].reject{|name, value| special_filters.include? name}
+    end
+    if filter_params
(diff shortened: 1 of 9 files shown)


app/datatables/datatable_base.rb
```diff
@@ -120,4 +123,8 @@ def select_query
   def count_query
     raise 'Must implement in subclass'
   end
+
+  def special_filters
+    raise 'Must implement in subclass'
```
acoffman, line 128: I don't know which is better, totally up to you, but you could have this return an empty array as the default behavior and then allow the two subclasses to override it with the parameters they care about. Perhaps less explicit, but also less duplicate methods. Both ways have their pros and cons.

app/datatables/datatable_base.rb
```diff
@@ -33,7 +33,10 @@ def paginate(objects)
   end
 
   def filter(objects)
-    if filter_params = params['filter']
+    if params['filter']
+      filter_params = params['filter'].reject{|name, value| special_filters.include? name}
```
acoffman, line 37: Not pushing a new commit for as its a tiny nitpick, but if you want you can use `_` in lieu of a variable name for variables you don't intend to use, that way if you run like a style checker or something it won't complain about declaring an unused variable.

susannasiebert: `filter_params` is used in the next if-clause.

acoffman: My bad, that was ambiguous, I meant `value` inside `reject` block.

susannasiebert: Doh. I'll remember that for the future.

PR #622, opened 2020-07-07
Update Sequence Ontology obo file location

susannasiebert:
The Sequence Ontology restructured their repo and changed the file names and location in the process. If someone could confirm that this is indeed the file we want to import, that would be great.


susannasiebert:
@malachig confirmed that all of the terms currently in use by CIViC appear in the updated file.

PR #625, opened 2020-07-17
Fix org stats to count correctly by Event org and not by users

susannasiebert:
Closes https://github.com/griffithlab/civic-client/issues/1433

Previously, we would take all of the users of an organization and add their stats together. Now that a user can be part of multiple orgs we need to actually look at the Event and only count the one attribute to this organization.

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -20,13 +20,23 @@ class Organization < ActiveRecord::Base
   end
 
   def stats_hash
-    (Hash.new { |h, k| h[k] = 0 }).tap do |accum|
-      all_users.each do |u|
-        u.stats_hash.each do |k, v|
-          accum[k] += v
-        end
-      end
-    end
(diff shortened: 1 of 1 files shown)


app/models/organization.rb
```diff
@@ -20,13 +20,23 @@ def self.datatable_scope
...
+      'submitted_evidence_items': e_ids.count,
+      'accepted_evidence_items': EvidenceItem.where(id: e_ids, status: 'accepted').count,
+      'suggested_sources': Event.where(organization_id: org_and_suborg_ids).where(action: 'publication suggested').count,
+      'submitted_assertions': a_ids.count,
+      'accepted_assertions': Assertion.where(id: a_ids, status: 'accepted').count,
+    }
+  end
+
+  def org_and_suborg_ids
+    return [self.id] + self.groups.map{|g| g.id}
```
acoffman, line 39: If you just want the primary key for a relation, there's a built-in helper `self.group_ids` will do what you want without loading the whole organization objects for the sub orgs.

PR #635, opened 2020-08-10
Check if a file exists before adding it to the releases hash

susannasiebert:


susannasiebert:
Do we? I thought that the frontend needs those nil values to determine whether or not to display buttons in the grid.

acoffman:
Ahh yes, I think you're right about that

PR #636, opened 2020-08-12
Variant coordinate fields should be NULLed out if unset

### LINKED ISSUE #606: Ensure that removed variant coordinate ref/alt are set to nil
When a variant has their reference and/or alt removed from their coordinates, that coordinate record now stores an empty string in those fields. We should ensure that these values get set to nil instead to keep those records in line with how they would be stored if no ref/alt had ever been associated with that variant.

acoffman:
For #606 we want to override the setter for certain attributes to replace empty strings with `NULL` for the purpose of saving it to the database. Unfortunately, we are already overriding the setter method for some of those attributes. Overriding it again will just replace one behavior with the other.

This PR does two things:

Introduces a new mixin `SetterOverrideMixin` which allows other mixins to safely stack setter override behavior. In other words, we can take a field and say, strip commas and whitespace, and also replace an empty string with null before saving. It accomplishes this by checking if we've already defined an overridden setter method. If so, it preserves an alias to the old method, defines the new one, then passes the normalized value along to the original method. In that way we can chain multiple normalizations together on the same attribute. There is an included test to demonstrate the functionality.

This also addresses issue 606 by marking all coordinate related fields of variant as attributes that should be nulled out if unset rather than saving an empty string.

closes #606

### DIFF
--- a/app/models/concerns/setter_override_mixin.rb
+++ b/app/models/concerns/setter_override_mixin.rb
@@ -1,0 +1,29 @@
+module SetterOverrideMixin
+
+  #this allows multiple overrides of the setter method
+  #to be safely combined by calling the old implementation
+  #and then the new one
+  #
+  #pass in the column to override, a unique name for your override,
+  #and a block containing the normalization logic for the value
+  def safe_override_setter(col, mixin_name, &block)
+    setter_name = "#{col}="
+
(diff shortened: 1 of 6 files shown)


acoffman:
This will work on the admin console yep! Yeah, I assume almost all fields we'll want this behavior for but I wasn't exactly sure

PR #640, opened 2020-09-08
Return 404 when an id isn't found

acoffman:
Currently when we call `find_by`, if the id doesn't exist, we get an error. Instead we should catch that error and return a 404 to be more idiomatic.

This will address errors in airbrake of the following style:

https://griffithlab-apps.airbrake.io/projects/285901/groups/2826758699422111844/notices/2826758673858709132
https://griffithlab-apps.airbrake.io/projects/285901/groups/2815961221013757690?tab=overview

etc.

Basically anything that's currently throwing an `ActiveRecord::NotFound` exception


susannasiebert (approved):
+1

PR #641, opened 2020-09-08
Handle case where /id/ matches but its not a valid id

acoffman:
This will address:

https://griffithlab-apps.airbrake.io/projects/285901/groups/2825207910287333077

Basically a shortlink that matches the `/id/` regex but doesn't actually resolve to a valid format (EID, AID, etc. The regex would return `nil` and then calling `.captures` on `nil` would error out.


susannasiebert (approved):
+1

PR #645, opened 2020-10-22
Add a flagged status column to flaggable entities

acoffman:
This introduces a new "flagged" status column for all flagged entities that will be kept up to date as flags are opened and resolved. Additionally it adds that status to the JSON responses for all flaggable entities.

closes griffithlab/civic-client#1513

### DIFF
--- a/db/migrate/20201022154530_add_flag_column.rb
+++ b/db/migrate/20201022154530_add_flag_column.rb
@@ -1,5 +1,6 @@
 class AddFlagColumn < ActiveRecord::Migration[5.2]
-  def change
+
+  def up
     [:evidence_items, :variants, :assertions, :variant_groups, :genes].each do |table|
       add_column table, :flagged, :boolean, null: false, index: true, default: false
     end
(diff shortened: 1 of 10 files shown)


db/migrate/20201022154530_add_flag_column.rb
```diff
@@ -0,0 +1,16 @@
+class AddFlagColumn < ActiveRecord::Migration[5.2]
+  def change
+    [:evidence_items, :variants, :assertions, :variant_groups, :genes].each do |table|
+      add_column table, :flagged, :boolean, null: false, index: true, default: false
+    end
+
+    [EvidenceItem, Variant, Assertion, VariantGroup, Gene].each do |klass|
```
susannasiebert, line 7: I don't think this will work on a rollback so you'll want to split this out in an `up` and `down` method.

susannasiebert (changes requested):
Looks good to me for the most part. Just one small suggestion for the migration file.

db/migrate/20201022154530_add_flag_column.rb
acoffman: Good catch!

susannasiebert (approved):
+1

db/structure.sql
```diff
@@ -11,6 +11,8 @@ SET row_security = off;
 
 SET default_tablespace = '';
 
+SET default_table_access_method = heap;
```
susannasiebert, line 14: I think we need to remove this line to make tests pass.

PR #651, opened 2020-11-02
Update MergeVariants utility to also move assertions and source suggestions

susannasiebert:
Closes https://github.com/griffithlab/civic-client/issues/1400


acoffman (approved):
+1

PR #652, opened 2020-11-02
Implement a job to update CIVIC's HPO records

susannasiebert:
We didn't have a job in place to update the HPO terms, which is why the Pediatric Onset term never got imported.

This new job grabs the HPO obo file and imports all term that are not deprecated. It updates existing term names and deletes terms that have become obsolete and have no evidence items or assertions that use the obsolete term. Any terms in active use would need to be manually fixed by making and accepting a suggested change to the affected evidence item/assertion. After them term has been removed from use the background job would take care of deleting the term next time it runs.

This job will run weekly.

Closes https://github.com/griffithlab/civic-client/issues/1417

### DIFF
--- a/lib/scrapers/human_phenotype_ontology.rb
+++ b/lib/scrapers/human_phenotype_ontology.rb
@@ -35,8 +35,20 @@ module Scrapers
           name = row.at_xpath('rdfs:label').text
           if !row.at_xpath('owl:deprecated').nil? && row.at_xpath('owl:deprecated').text == 'true'
             p = Phenotype.find_by(hpo_id: hpo_id)
-            if !p.nil? && p.evidence_items.count == 0 && p.assertions.count == 0
-              p.delete
+            if !p.nil?
+              if p.evidence_items.count == 0 && p.assertions.count == 0
+                p.delete
+              else
+                civicbot_user = User.find(385)
+                (p.evidence_items + p.assertions).each do |obj|
+                  if obj.flags.select{|f| f.state == 'open' && f.comments.select{|c| c.title == "Deprecated HPO term" && c.user_id = 385}.count > 0}.count == 0
+                    result = Flag.create_for_flaggable(civicbot_user, obj, nil)
+                    if result.succeeded?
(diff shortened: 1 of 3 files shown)


lib/scrapers/human_phenotype_ontology.rb
```diff
@@ -24,6 +24,41 @@ def self.scrape_all()
...
+            if !p.nil?
+              if p.evidence_items.count == 0 && p.assertions.count == 0
+                p.delete
+              else
+                civicbot_user = User.find(385)
+                (p.evidence_items + p.assertions).each do |obj|
+                  if obj.flags.select{|f| f.state == 'open' && f.comments.select{|c| c.title == "Deprecated HPO term" && c.user_id = 385}.count > 0}.count == 0
+                    result = Flag.create_for_flaggable(civicbot_user, obj, nil)
+                    if result.succeeded?
+                      Comment.create({title: 'Deprecated HPO term', text: "This entity uses a deprecated HPO term \"#{name}\" (#{hpo_id})", user: civicbot_user, commentable: result.flag})
```
acoffman, line 47: You don't need to nest this all in a hash, `Comment.create(title: 'foo', text: 'bar'... etc)` is sufficient.

lib/scrapers/human_phenotype_ontology.rb
```diff
@@ -24,6 +24,41 @@ def self.scrape_all()
...
+          hpo_id = hpo_id.sub('_', ':')
+          name = row.at_xpath('rdfs:label').text
+          if !row.at_xpath('owl:deprecated').nil? && row.at_xpath('owl:deprecated').text == 'true'
+            p = Phenotype.find_by(hpo_id: hpo_id)
+            if !p.nil?
+              if p.evidence_items.count == 0 && p.assertions.count == 0
+                p.delete
+              else
+                civicbot_user = User.find(385)
+                (p.evidence_items + p.assertions).each do |obj|
```
acoffman, line 43: We could probably preload flags and their comments here to make things a little more efficient, but at the moment I imagine its a non-issue.

acoffman (approved):
Looks good overall!

PR #653, opened 2020-11-05
Fix naming errors when resolving flags

acoffman:
closes griffithlab/civic-client#1535


susannasiebert (approved):
+1 I haven't tested this locally but this looks correct to me.

Issue #650, opened 2020-10-31
Do not allow entry of obsolete DOID terms


ChR-1610:
Currently variants can be found that are annotated with obsolete DOID terms. For the example the term 'Anaplastic Thyroid Cancer' is now obsolete (DOID:7012) and has in the ontology been replaced by the term 'Thyroid Gland Anaplastic Cancer' (DOID:0080522).
Ensure obsolete disease ontology terms cannot be used.

PR #654, opened 2020-11-11
Flag evidence items and assertions that use DO terms that are obsolete, not in the cancer slim file, or are missing DOIDs

susannasiebert:
This PR does a few things:
- It deletes disease aliases that are no longer in the DO
- It deletes disease terms that are not in the cancer slim file and also not used by any evidence items or assertions
- It flags evidence items and assertions that use a DO term that is not in the cancer slim file (so that we can request those terms be added to the slim file)
- It flags evidence items and assertions that use a DO term that has become obsolete (so that we can merge the disease with the term that supersedes the obsolete term)
- It flags evidence items and assertions that use a disease with a DOID in the CIViC database (so that we can add the DOID in the admin interface, have the term added to the slim file and/or have it merged with a better term already available)

It also fixes some URLs that have been updated.

This should address https://github.com/griffithlab/civic-client/issues/1359, https://github.com/griffithlab/civic-client/issues/1470, https://github.com/griffithlab/civic-client/issues/1469, and https://github.com/griffithlab/civic-server/issues/650

There are are already about 150 disease terms identified that have one of the issues described above and would result in flagging once this PR is merged and the job runs the next time. In the interest of preventing a bunch of flags being created, we will probably want to resolve those before merging this PR.


acoffman (approved):
LGTM

susannasiebert:
Also closes https://github.com/griffithlab/civic-client/issues/1509

susannasiebert:
And also closes https://github.com/griffithlab/civic-client/issues/1270

(I think this officially the PR that closes the most issues)

susannasiebert:
I checked this morning the DO cancer slim file has not yet been re-created after adding the new terms. So this PR needs to stay open a while longer. According to this comment (https://github.com/DiseaseOntology/HumanDiseaseOntology/issues/920#issuecomment-739008323) from Lynn, it should be any day now though.

susannasiebert:
An updated DO obo file is now available that has the terms we asked to have added. This PR can be merged.

PR #655, opened 2020-11-12
Add flagged status to datatable results

acoffman:
This PR adds the `flagged` attribute to the datatable presenters and queries.


susannasiebert (approved):
+1

PR #656, opened 2020-11-12
Include updated_at timestamp in index responses 

acoffman:


susannasiebert (approved):
+1 I do wonder if it would be more helpful for them to have a new endpoint that returns items that were updated in a specified timeframe.

PR #658, opened 2020-12-03
Fix column name

acoffman:
When sorting the variant group data table by name we'd get an error as the
column name was incorrect. This corrects it.

https://griffithlab-apps.airbrake.io/projects/285901/groups/2860600362368029143/notices/2860600354194289502


susannasiebert (approved):
+1

PR #660, opened 2020-12-03
Admin eid fix

acoffman:
Two small things:

* When doing foreground processing the arg is passed to perform rather than the constructor
* Keep "eids requiring action" from loading everything so the admin dashboard doesn't take forever to load


susannasiebert (approved):
+1

PR #664, opened 2020-12-09
Add Oncogenic Clinical Significance

susannasiebert:
Prerequisite for https://github.com/griffithlab/civic-client/issues/1455

### DIFF
--- a/app/models/actions/propose_evidence_item.rb
+++ b/app/models/actions/propose_evidence_item.rb
@@ -49,7 +49,7 @@ module Actions
       if params[:noDoid]
         proposed_name = Disease.capitalize_name(params[:disease_name])
         Disease.where(display_name: proposed_name, name: proposed_name).first_or_create
-      else
+      elsif params[:disease]
         Disease.find_by(id: params[:disease][:id])
       end
     end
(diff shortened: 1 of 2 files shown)


jmcmichael:
Adding Functional, non-Oncogenic Evidence appears to return a `400: undefined method 'strip' for nil:NilClass` error from the server.

jmcmichael:
The 400 error was due to the disease fields improperly cleared 'noDoid' field, fixed in the latest griffithlab/civic-client#1570 PR commits.

app/models/actions/propose_evidence_item.rb
```diff
@@ -49,7 +49,7 @@ def get_disease(params)
       if params[:noDoid]
         proposed_name = Disease.capitalize_name(params[:disease_name])
         Disease.where(display_name: proposed_name, name: proposed_name).first_or_create
-      else
+      elsif params[:disease]
```
acoffman, line 52: This isn't a problem for this PR to fix, more a comment on our general need for better validation around EID state, but now there's no distinction between submitting a valid EID that just doesn't have a disease because of its type, and submitting an EID that _should_ have a disease and just not supplying the parameter.

For V2 I think we should just rework this logic with more formal validation across the board.

acoffman (approved):

PR #665, opened 2020-12-10
Skip import of and delete existing obsolete sequence ontology terms and flag variants that use such terms

susannasiebert:
Closes https://github.com/griffithlab/civic-client/issues/1558

### DIFF
--- a/lib/importer/sequence_ontology_mirror.rb
+++ b/lib/importer/sequence_ontology_mirror.rb
@@ -37,10 +37,9 @@ module Importer
     def valid_entry?(entry)
       if entry['is_obsolete'].present?
         obsolete_terms.append(entry)
+        return false
       else
-        ['id', 'name'].inject(true) do |val, term|
-          entry[term].present? && val
-        end
+        ['id', 'name'].all? { |term| entry[term].present? }
       end
     end
 
@@ -103,7 +102,6 @@ module Importer
             if term['replaced_by'].present?
               text += " Replaced by #{term['replaced_by']}."
             end
-            binding.pry
             obsolete_type.variants.each do |variant|
               if variant.flags.select{|f| f.state == 'open' && f.comments.select{|c| c.title == title && c.user_id = 385}.count > 0}.count == 0
                 result = Flag.create_for_flaggable(civicbot_user, variant, nil)
--- a/lib/importer/sequence_ontology_mirror.rb
+++ b/lib/importer/sequence_ontology_mirror.rb
@@ -1,10 +1,11 @@
 module Importer
   class SequenceOntologyMirror
-    attr_reader :parser, :version
+    attr_reader :parser, :version, :obsolete_terms
 
     def initialize(path, version = Time.now.utc.iso8601)
       @parser = Obo::Parser.new(path)
-      @verison = version
+      @version = version
+      @obsolete_terms = []
     end
 
     def import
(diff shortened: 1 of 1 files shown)


lib/importer/sequence_ontology_mirror.rb
```diff
@@ -33,8 +35,12 @@ def populate_ontology_entry
     end
 
     def valid_entry?(entry)
-      ['id', 'name'].inject(true) do |val, term|
-        entry[term].present? && val
+      if entry['is_obsolete'].present?
+        obsolete_terms.append(entry)
+      else
+        ['id', 'name'].inject(true) do |val, term|
```
acoffman, line 41: Dunno if it reads as any clearer to you, but this could also be `['id', 'name'].all? { |term| entry[term].present? }`what you've got will work just fine as well though. (It looks like you didn't actually make that change in the first place, its just formatting so don't feel like you have to change it)

lib/importer/sequence_ontology_mirror.rb
```diff
@@ -80,5 +86,35 @@ def create_parent_links
...
+            civicbot_user = User.find(385)
+            title = "Obsolete SO Term"
+            text = "This variant uses an obsolete Sequence Ontology term #{obsolete_type.display_name} (#{obsolete_type.soid})."
+            if term['consider'].present?
+              text += " Consider #{term['consider']}."
+            end
+            if term['replaced_by'].present?
+              text += " Replaced by #{term['replaced_by']}."
+            end
+            binding.pry
```
acoffman, line 106: breakpoint left in

lib/importer/sequence_ontology_mirror.rb
```diff
@@ -33,8 +35,12 @@ def populate_ontology_entry
     end
 
     def valid_entry?(entry)
-      ['id', 'name'].inject(true) do |val, term|
-        entry[term].present? && val
+      if entry['is_obsolete'].present?
+        obsolete_terms.append(entry)
```
acoffman, line 39: `append` will return a copy of the array, which will be a truthy value, which means `valid_entry?` will report as true and the term will be added, only to be removed when we `process_obsolete_terms`. Is that intentional? Or would it make sense to return something false here and not add them in the first place if they're not already present?

acoffman (changes requested):

lib/importer/sequence_ontology_mirror.rb
susannasiebert, line 39:
```suggestion
        obsolete_terms.append(entry)
        return false
```

susannasiebert: I didn't realize it would return a truthy value. I added a line to `return false`

lib/importer/sequence_ontology_mirror.rb
```diff
@@ -33,8 +35,12 @@ def populate_ontology_entry
...
+      else
+        ['id', 'name'].inject(true) do |val, term|
+          entry[term].present? && val
+        end
```
susannasiebert, line 43:
```suggestion
        ['id', 'name'].all? { |term| entry[term].present? }
```

lib/importer/sequence_ontology_mirror.rb
susannasiebert, line 106:
```suggestion
```

susannasiebert (commented):
@acoffman I made a few code suggestions for your comments. Could you have a quick look at them and confirm that they look correct?

acoffman:
They all look good to me!

acoffman (approved):

PR #667, opened 2021-02-03
Recalculate the allele registry id after a Variant is updated

susannasiebert:
Closes #1474

### DIFF
--- a/app/models/variant.rb
+++ b/app/models/variant.rb
@@ -44,7 +44,7 @@ class Variant < ActiveRecord::Base
   end
 
   def update_allele_registry_id
-    SetAlleleRegistryIdSingleVariant.new().perform(self)
+    SetAlleleRegistryIdSingleVariant.perform_later(self)
   end
 
   def self.index_scope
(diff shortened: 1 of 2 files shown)


app/models/variant.rb
```diff
@@ -42,6 +43,10 @@ def init
     self.civic_actionability_score ||= 0 if self.has_attribute? :civic_actionability_score
   end
 
+  def update_allele_registry_id
+    SetAlleleRegistryIdSingleVariant.new().perform(self)
```
acoffman, line 47: I believe you'll want this to be `SetAlleleRegistryIdSingleVariant.perform_later(self)` otherwise it will execute synchronously rather getting queued in the background.

acoffman (changes requested):
One small change but then good to merge; looks good!

app/models/variant.rb
susannasiebert: Fixed.

acoffman (approved):
LGTM

PR #668, opened 2021-02-04
Defer to original implementation on nil

acoffman:
When an object is `nil` just use the original assignment rather than the overridden one.


susannasiebert (approved):
+1

PR #671, opened 2021-02-08
Fix logic of when to set the allele registry id

susannasiebert:
Previously, the check would not catch the case where the allele registry id came back as `_:CA`. We would then set the allele registry id again to 'unregistered' which would trigger another round of this job getting called, creating an infinite loop.


acoffman (approved):
+1

PR #672, opened 2021-02-09
Add logic check so _:CA allele_registry_ids don't get processed in second clause

susannasiebert:
#671 still wasn't quite right.


acoffman (approved):
+1

PR #673, opened 2021-02-12
Some improvements to the UpdateClinicalTrials job

susannasiebert:


app/jobs/update_clinical_trials.rb
```diff
@@ -1,14 +1,19 @@
...
   def perform
-    Source.where(source_type: 'PubMed').each do |source|
+    Source.eager_load(:clinical_trials).where(source_type: 'PubMed').each do |source|
       resp = Scrapers::PubMed.call_pubmed_api(source.citation_id)
       clinical_trials = resp.clinical_trial_ids.uniq.map do |nct_id|
         ClinicalTrial.where(nct_id: nct_id).first_or_create
       end
-      source.clinical_trials = clinical_trials
-      source.save
+      if clinical_trials.sort != source.clinical_trials.sort
```
acoffman, line 9: Arrays do have built in union, intersect, and difference methods which you could use here if you  thought it was more clear, but this works just fine!

acoffman (approved):
+1

PR #674, opened 2021-02-15
Fix code that assumes that the disease field is required on evidence items

susannasiebert:
Functional evidence items no longer allow a disease to be set but in different parts of the codebase we assume that all evidence items will have a disease assigned to it. This PR fixes the places I found where that is the case.

Closes https://github.com/griffithlab/civic-client/issues/1589


app/presenters/evidence_item_index_presenter.rb
```diff
@@ -10,7 +10,7 @@ def as_json(opts = {})
         id: item.id,
         name: item.name,
         description: item.description,
-        disease: DiseasePresenter.new(item.disease),
+        disease: item.disease.nil? ? nil: DiseasePresenter.new(item.disease),
```
acoffman, line 13: I wonder if it would be cleaner to encode this logic in the disease presenter itself so you don't have to duplicate it here and in the grid presenter both (and potentially elsewhere). In the presenter you could just do something like

```
def as_json

  if disease.nil?
     nil
  else
    # existing code
   end
end
```

and then anywhere we used a disease presenter it could handle it being nil.

acoffman (approved):
Looks good as is, maybe one small improvement but I don't have a strong feeling either way.

app/presenters/evidence_item_index_presenter.rb
susannasiebert: I like this suggestion as a fallback but it feels weird to me to set it up in a way that the disease presenter itself _should_ have to worry about it.

Luckily with V2, this all becomes moot anyway.

PR #675, opened 2021-02-18
Add additional disease terms that need special handling

susannasiebert:
This PR adds a few DOIDs that are not in the cancer slim file and do not belong there but we still want to support them in CIViC.

It also handles a new class of disease terms that the DO does not wish to include in their ontology and so will never have a DOID.


acoffman (approved):
LGTM

PR #681, opened 2021-03-29
Use the evidence_items_by_status view in variant search

acoffman:
In some cases, placing the conditional in a HAVING clause and grouping
was not having the desired effect. This is more straightforward to
reason about (and also works consistently).

Note that the view is already scoped to only include non-deleted EIDs so
the behavior there will reamin the same despite the removal of the
previous where clause.


susannasiebert (approved):
+1

PR #682, opened 2021-04-08
Fill in full_journal_title for ASCO abstracts

acoffman:
This value was not previously set, and it is now hardcoded for ASCO sources. We have backfilled this for existing sources, and this should fix the issue going forward.

We will also want to regenerate clinical trial associations.

closes griffithlab/civic-client#1612


susannasiebert (approved):
+1

PR #683, opened 2021-04-09
Handle searching by combination type is not null correctly

acoffman:
Previously it would make a sub select of all EIDs with multiple drugs but then evaluate that the eid was not in that set, which was not the desired behavior.

closes griffithlab/civic-client#1581


susannasiebert (approved):
+1

PR #688, opened 2021-06-07
Gracefully handle ORCID login when user hasn't specified a name

acoffman:
Something seems to have changed in their API responses. This field was previously always present but in some cases it no longer is. 

In the case when we have no family or given name, we'll fall back on the orcid itself as a new user's name.


susannasiebert (approved):
+1

PR #689, opened 2021-06-07
Update Rails minor version

acoffman:
This updates the rails minor version to address the dependabot alert.


susannasiebert (approved):
+1

PR #691, opened 2021-09-02
Create UniProt Mapping file. 

nairod2000:
Corrected spelling error in generate_tsvs.rb, created tsv generator and presenter, scheduled the task to run monthly, added functions to get uniprot names and created the uniprotmapping tsv. Need help getting the correct counts for variants and evidence items that are non-rejected.


app/jobs/generate_uniprot_mapping_tsv.rb
```diff
@@ -0,0 +1,54 @@
+class GenerateUniprotMappingTsv < GenerateTsvs
+    def perform
+      ensure_downloads_directory_exists
+      tsvs_to_generate.each do |e|
+        begin
+          tmp_file = tmp_file(e.file_name)
+          tmp_file.puts(e.headers.join("\t"))
+  
+          e.objects.find_each do |object|
```
acoffman, line 9: You probably don't need the indirection here of calling `e.objects` and can delete `def self.objects` in the presenter file. This can probably just be `Gene.find_each`

app/jobs/generate_uniprot_mapping_tsv.rb
```diff
@@ -0,0 +1,54 @@
...
+            
+            row = e.row_from_object(object)
+            if row[1].is_a?(Array)
```
acoffman, line 12: I might change this to be `e.rows_from_object()` and assume you always get back an Array of rows and push the logic of handling multiple (or no) uniprot ids down a level. (See other comment)

app/presenters/uniprot_mapping_tsv_presenter.rb
```diff
@@ -0,0 +1,34 @@
...
+  
+    def self.headers
+      [
+        'civic_name',
+        'uniprot_name',
+        'gene_overview'
+      ]
+    end 
+  
+    def self.row_from_object(gene)
```
acoffman, line 14: I'd rename to `rows_from_object()` and do something along these lines (haven't tested it, just off the top of my head):
```
swissprot_names = Array(Scrapers::MyGeneInfo.get_swissprot_name(gene))
formatted_overview = formatted_overview_col(gene)
swisprot_names.map do |swissprot_name|
  if name == 'N/A'
    nil
  else 
    [gene.name, swissprot_name, formatted_overview]
  end
end.compact
```

That way you have a list of rows for your TSV, `compact` will remove the `nil`s and the code that actually writes the TSV can just be a simple iteration over genes, calling this, and then writing a row for each item this returns.

app/presenters/uniprot_mapping_tsv_presenter.rb
```diff
@@ -0,0 +1,34 @@
...
+        Scrapers::MyGeneInfo.get_swissprot_name(gene),
+        formatted_overview_col(gene)
+      ]
+    end
+  
+    def self.file_name
+      "UniprotMapping.tsv"
+    end 
+  
+    def self.formatted_overview_col(gene)
```
acoffman, line 26: I'd get the counts in the following ways:

```
eid_count = EvidenceItem.joins(variant: [:gene]).where("evidence_items.status != 'rejected'").where(variant: {gene: gene}).distinct.count
variant_count = gene.variants.joins(:evidence_items).where("evidence_items.status != 'rejected'").distinct.count
assertion_count = gene.assertions.where("status != 'rejected'").distinct.count
```

You could also invert the logic and do something like this:

```
Assertion.joins(:gene).where("status != 'rejected'").where(gene: g).distinct.count
```
depending on what's more clear to you.

acoffman (changes requested):
Made some comments based on what we talked about on the call - let me know if you have any questions!

Issue #701, opened 2022-04-01
Dockerize civic application


forus:
Hi there, I could not find a Dockerfile in this project nor a repository that would build docker images in the git organisation.

Is it already implemented somewhere? If not, would contributors be interested to accept such change?

pieterlukasse:
Hi @acoffman,
any pointers on this one? We can start on it soon, if there is nothing else out there. Please let us know!
CC @malachig

acoffman:
Hi guys, thanks for reaching out and apologies for the delayed response. You are correct that there is no Dockerfile for this project. We would be happy to accept such a change, however, you've caught us in a bit of transitional moment as we are in the final sprint to launching V2 of CIViC which is a new codebase (https://github.com/griffithlab/civic-v2). 

While we intend to keep the v1 API alive for some time for backwards compatibility, eventually, the new version will entirely replace this repo. Given that, it may be more productive to Dockerize the new codebase. 

Its a fairly standard rails deployment with nginx proxying requests to puma over a unix socket, sidekiq used for a task queue and redis used for the queue itself, with postgres as the primary data store. Systemd unit files and nginx config are in the new repo, as are github action workflow that build the frontend and deploy the whole application, which should be a good starting point. 

If this is something you're interested in tackling, I can also share notes on the dependencies you'll likely need.

pieterlukasse:
@acoffman, @susannasiebert thanks for sharing all these details. Good to hear that this project is doing so well and that a new version is coming out! 

We are still interested in the v1, given that the application we want to integrate this with is expecting v1 version of the API. Having said that, my question to @acoffman is: how different is the v2 API? I might have to start planning a migration of our code at some point. Can you please also let us know for how long the v1 will be supported with new data releases? There is a chance that our use case will require us to update our local instance of the Civic DB once every ~3 months or so. If v1 is sunsetting soon, it would be nice to know so we can take that into account.

@susannasiebert given our use case (having a local copy of the Civic DB), can you please help us find a recent dump of the DB that we can use to populate our instance? I found this one, but it is >1 year old: https://github.com/griffithlab/civic-server/blob/staging/db/data.sql.gz  ... and if I understand the README correctly, it is just a "sanitized version" for dev purposes. Where can we find a recent dump? Thanks!

acoffman:
[...]

PR #704, opened 2022-05-10
Dockerize civic-server

ruslan-forostianov:
Please see https://github.com/griffithlab/civic-server/issues/701

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -36,12 +36,15 @@ Rails.application.configure do
 
   # `config.assets.precompile` has moved to config/initializers/assets.rb
 
+  #run background tasks syncronously
+  config.active_job.queue_adapter = :inline
+
   # Specifies the header that your server uses for sending files.
   # config.action_dispatch.x_sendfile_header = "X-Sendfile" # for apache
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for nginx
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = true
+  #config.force_ssl = true
 
   # Set to :debug to see everything in the log.
   config.log_level = :info
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -17,9 +17,8 @@ else
     echo "Creating civic database"
     rake db:create
 
-    echo "Migrating civic database"
-    # migrate:
-    rake db:migrate
+    echo "Initializing civic database"
+    rake db:structure:load
 
     echo "Loading a recent database backup"
     # unpack ./db/data.sql.gz (required before loading data dump):
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -4,30 +4,29 @@ set -e
 set -u # unset variables throw error
 set -o pipefail # pipes fail when partial command fails
 
-bundle update --bundler 
-# make sure env $DATABASE_* items are set!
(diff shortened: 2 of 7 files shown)


config/environments/production.rb
```diff
@@ -36,12 +36,15 @@
...
+  #run background tasks syncronously
+  config.active_job.queue_adapter = :inline
+
   # Specifies the header that your server uses for sending files.
   # config.action_dispatch.x_sendfile_header = "X-Sendfile" # for apache
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for nginx
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = true
+  #config.force_ssl = true
```
pieterlukasse, line 47: this was needed because our deployment is behind reverse proxy

config/environments/production.rb
```diff
@@ -36,12 +36,15 @@
 
   # `config.assets.precompile` has moved to config/initializers/assets.rb
 
+  #run background tasks syncronously
+  config.active_job.queue_adapter = :inline
```
pieterlukasse, line 40: for some reason, if this is not set, the deployment returns HTTP 500 for any api call...not sure why.

pieterlukasse: Maybe @acoffman can comment on this?

acoffman: This is likely due to the default production background jobs queue requiring redis in production. Setting it to inline will cause any background tasks (delivering notifications, analytics, generating data dumps, etc) to trigger during the request/response cycle rather than being queued for later processing. 

For a smaller installation, that's probably fine, but if you notice performance issues, that could be a culprit.

docker-entrypoint.sh
```diff
@@ -0,0 +1,33 @@
...
+# db creation should be executed:
+if PGPASSWORD=$DATABASE_PASSWORD psql -lqt -U$DATABASE_USER -h$DATABASE_HOST | cut -d \| -f 1 | grep -q civic; then
+    echo "civic database already exists. Skipping database initialisation."
+else
+    echo "Creating civic database"
+    rake db:create
+
+    echo "Migrating civic database"
+    # migrate:
+    rake db:migrate
```
acoffman, line 22: Since you're initializing the database from scratch here, and not running migrations in the case of an update anyways, you could just run `rake db:structure:load` here which will load in the schema in its entirety at once rather than stepping through each migration.

acoffman (approved):
This looks good to me overall. There are a couple things (like an explicit puma configuration) that you might find you need to tweak as you see what the usage looks like in your deployment, but this is a great start. Happy to merge as is, and feel free to follow up with us if you run into any issues with your deployment.

docker-entrypoint.sh
ruslan-forostianov: @acoffman I've started using the command here https://github.com/griffithlab/civic-server/pull/704/commits/63d762cb3b1d5ce1977f9cbbde4d62759b5a5907

ruslan-forostianov:
> This looks good to me overall. There are a couple things (like an explicit puma configuration) that you might find you need to tweak as you see what the usage looks like in your deployment, but this is a great start. Happy to merge as is, and feel free to follow up with us if you run into any issues with your deployment.

Thanks, @acoffman !

pieterlukasse:
Awesome! Thanks @ruslan-forostianov and @acoffman ! 👍

